Issue #3, opened 2020-03-16
Define structure for regions


danielhuppmann:
This should provide instruction on how to define region names consistently, including a naming convention for overlapping regions (`EU`, `EU + UK`, `EU + UK, NO, CH`, `EU excluding MT, CY`, ...).

See the `CL_AREA` at the [SDMX global registry](https://registry.sdmx.org/items/codelist.html) for a template.

Region levels to be defined:
 - [x] aggregate regions
 - [x] list of countries
 - [x] sub-country specifications following the NUTSx classification
 - [x] other required spatial definitions (city, ...)

Required explanation:
 - [x] how to deal with directional data

sandrinecharousset:
I copy here the discussion about sub-country levels, which can be found 
- in a [commit in my fork where I edited the readme](https://github.com/sandrinecharousset/nomenclature/commit/703e26ddaca98bffd3c9514488d3319e73e0ac63)
- in #12 (but was at the wrong place)

danielhuppmann:
Thank you @sandrinecharousset for raising this issue. I suggest the following approach:

Add a file `subcountry-regions.yaml` following the implementation of `aggregate-regions.yaml`, where each entry has the following structure:

```
<name of region>:  # should be composed of the country name plus a meaningful identifier
   country: <parent country>  # must be a member of `countries.yaml`
   # optional attributes to be used as appropriate
   definition: <list of NUTS-x codes>  # must be a composed of members of `nuts[1/2/3].yaml`
   notes: <any relevant information, e.g., reference to another project or model documentation>
   source: <link to a document specifying this region>
```

I would not introduce any information about which subregions add up to the total country, because this could quickly become ambiguous, example based on @sandrinecharousset's comment:

 - one model uses `IT_N + IT_S`
 - another model uses `52_IT + 53_IT + IT_S`

So I would leave the responsibility of providing correct sub-country data to the modelling team.

PR #10, opened 2020-03-18
Adding aggregate-region and country specifications

danielhuppmann:
This PR adds a subfolder for the `region` dimension of the data format, specifying the convention for "aggregate-regions" and country-level data (NUTSx and custom specifications to come later).

The region-folder README also includes a description how to deal with directional data. 

The mapping of countries includes iso2- and iso3-codes as well as a flag for EU membership and alternative names used for some countries. The `countries.yaml` file is created from a data table and a Python script in the `data` subfolder.

implementing first part of #3


region/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+list_of_countries = list(country_mapping.keys())
+iso2_mapping = dict([(country_mapping[c]['iso2'], c)
+                     for c in country_mapping.keys()])
+```
+
+### Sub-country areas
+
+The disaggregation of countries follow the 
+[NUTS 2021 classification](https://ec.europa.eu/eurostat/web/nuts/background)
```
danielhuppmann: @erikfilias - see here for reference to NUTSx

danielhuppmann:
> @danielhuppmann, What do you think about if we adopt the use of codes from [NUTS 2021 classification](https://ec.europa.eu/eurostat/web/nuts/background) to specify zones/areas inside of a country?

Yes, the NUTSx codelists will be added in a subsequent PR - didn't want to put too much stuff into one PR because that would make it (even more) difficult to review. Added an inline comment to reference where this is included in the README.

region/README.md
erikfilias: I got it. That's great. Thanks.

PR #12, opened 2020-03-23
Add structure of variable section and import nomenclature from IAMC 1.5°C

### LINKED ISSUE #4: Define structure for variables & units and import nomenclature from IAMC 1.5°C
This should develop a structure for defining variables and units, as well as detailed instructions on how to suggest/add new ones.

As starting point, the relevant nomenclature used in the IPCC SR15 (see [the docs of the IAMC 1.5°C Scenario Explorer](https://data.ene.iiasa.ac.at/iamc-1.5c-explorer/#/docs)).

danielhuppmann:
This PR adds the structure for the `variable` and `unit` dimensions. It also creates a first set of codelists (lists of variables) from the IPCC SR15.

**To-dos:**
 - [x] define structure for variables and draft READMEs in subfolders
 - [x] import relevant nomenclature used for the IPCC SR15, see [the docs of the IAMC 1.5°C Scenario Explorer](https://data.ene.iiasa.ac.at/iamc-1.5c-explorer/#/docs)
 - [x] add raw list of variables from IAMC 1.5°C scenario ensemble to `variable/data` subfolder
 - [x] add processing scripts used to generate `yaml` files in this PR
 - [x] define structure for units

closes #4

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ In the data format, every timeseries is described by six dimensions (codes):
 1.	Model - [more information](model)
 2.	Scenario - [more information](scenario)
 3.	Region - [more information](region)
-4.	Variable
+4.	Variable - [more information](variable)
 5.	Unit
 6.	Subannual (optional, default 'Year')<sup>[1]</sup>
 
--- a/README.md
+++ b/README.md
@@ -50,14 +50,14 @@ In the data format, every timeseries is described by six dimensions (codes):
 2.	Scenario - [more information](scenario)
 3.	Region - [more information](region)
 4.	Variable - [more information](variable)
-5.	Unit
-6.	Subannual (optional, default 'Year')<sup>[1]</sup>
-
-**Work in progress**: A detailed description of these dimensions and the shared
-terms (i.e., codelists) will be added soon!
+5.	Unit - see the section on [variables](variable) for details
+6.	Subannual (optional, default 'Year')<sup>[1][2]</sup>
 
 [1] *The index 'Subannual' is an extension of the original format introduced by
 the openENTRANCE project to accomodate data at a subannual temporal resolution.*
+
+[2] *A detailed description of this dimensions and the shared terms
+(i.e., codelists) will be added soon!**
 
 ### Recommended usage of this data format
 
--- a/region/countries.yaml
+++ b/region/countries.yaml
@@ -1,4 +1,4 @@
-# This file was created using the script `utils/write-countries.py`
+# This file was created using the script `data/write-countries.py`
 # DO NOT ALTER THIS FILE MANUALLY!
 
 # List of countries
--- a/region/data/write-countries.py
+++ b/region/data/write-countries.py
@@ -8,7 +8,7 @@ countries = pd.read_csv('countries.csv', header=None)
 
 # open file and write header
 file = open('../countries.yaml', 'w')
-file.writelines(f'# This file was created using the script `utils/{this}`\n')
+file.writelines(f'# This file was created using the script `data/{this}`\n')
 
 file.write('# DO NOT ALTER THIS FILE MANUALLY!\n\n')
 file.write('# List of countries\n')
--- a/variable/README.md
+++ b/variable/README.md
@@ -1,0 +1,112 @@
+# Guidelines for variable definitions
+
+The **variable** column specifies which type of information is provided in
+the specific timeseries. This can be for example quantities of energy, prices,
+economic activity, or specifications of technology parameters.
+
+## Overview
+
+### The variable hierarchy
+
+This column implements a "semi-hierarchical" structure using the `|` character
+(pipe, not l or i) to indicate the *depth*.
+Semi-hierarchical means that a hierarchy can be imposed,
+e.g., a user can enforce that the sum of `Emissions|CO2|Energy` and
+`Emissions|CO2|Other` must be equal to `Emissions|CO2`
+(if there are no other `Emissions|CO2|…` variables).
+However, this is not mandatory, e.g., the sum of `Primary Energy|Coal`,
+`Primary Energy|Gas`, `Primary Energy|Fossil` and other `Primary Energy|…`
+data should not be expected to equal `Primary Energy`
+because this would double-count fossil fuels.
+
+### Naming convention for variables
+
+When suggesting/adding new variables, please follow these rules:
+
+- A `|` (pipe) character indicates levels of hierarchy
+- Do not use spaces before and after the `|` character,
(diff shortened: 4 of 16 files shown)


sandrinecharousset:
The structure of folders looks relevant.
the readme's and yaml look sufficient and are understandable even by a non specialist such as myself.

sandrinecharousset:
Units in the description of variables : a unit is mentioned for all variables, where it may appen than different units can be widely used by different communities. One example is energy. In the current description, the unit is EJ (hexaJoule), while usually energy modellers (from the field I am attached to) use MW/MWh/GW/GWh (MegaWatt, MegaWatt-hour, GigaWatt, GigaWatt-hour). There exist an easy conversion (1Wh=3600J) but the magnitudes of main variables are well known by modellers, with their unit only (energy in EJ doesn't speak to me while energy in GWh does). Can we link a serie of recommended units to each variable?  Knowing that in the end the unit is in each line of the data.

danielhuppmann:
thank you for the good comment @sandrinecharousset!

> Can we link a serie of recommended units to each variable?

There are several options:

1. allow a list of (recommended) units for each variable (as long as they are the same type, e.g., energy)
   1. store the data as uploaded by the teams on the platform (potentially making it difficult to compare across models)
   2. convert units during upload to one standard unit (meaning that data displayed and downloaded will have a different unit that what is uploaded by the modelling team)
2. only allow a specific unit for each variable

We plan to have a feature for unit-conversion in the Scenario Explorer display, but this will probably not be ready before autumn 2020.

sandrinecharousset:
> > Can we link a serie of recommended units to each variable?
> 
> There are several options:
> 
> 1. allow a list of (recommended) units for each variable (as long as they are the same type, e.g., energy)
>    1. store the data as uploaded by the teams on the platform (potentially making it difficult to compare across models)
>    2. convert units during upload to one standard unit (meaning that data displayed and downloaded will have a different unit that what is uploaded by the modelling team)

> 2. only allow a specific unit for each variable
=> but how to decide which one?
> 
> We plan to have a feature for unit-conversion in the Scenario Explorer display, but this will probably not be ready before autumn 2020.

=>My personal feeling is that whenever it is possible to get a compromise on units we should target to have one onIy but it may be difficult not to keep a (short) list of variables with 2 units ( eg the EJ vs GWh). Probably when it is about units that are very close (GW/MW) it will be easy to chose only one, but when it is units that may change the understanding of the data (which shouldn't be numerous), it may be impossible.... The unit-conversion feature above proably solves all this, so if it has to be only temporary maybe we can work with 1 unit for all variables (but which one?)

sandrinecharousset:
About the regions definition: from the readme seems we cannot have intermediates between countries and nuts1 regions. in plan4res we use for one country (italy) aggregates which are inbetween (italy separated in ITN and ITS, for noth italy and south italy). Can we give the possibility of having other regionalisations than nuts?

arght:
Sandrine,
About regions it depends how flexible we want to be. Allowing intermediate regions introduces flexibility but again may introduce difficulty in the use of the data set, because there can be several definitions of what it Italy North and Italy South. Alternatively, we can use just two NUTS1 from Italy to represent what you want Italy North and Italy South
Best regards
Andrés

sandrinecharousset:
> Sandrine,
> About regions it depends how flexible we want to be. Allowing intermediate regions introduces flexibility but again may introduce difficulty in the use of the data set, because there can be several definitions of what it Italy North and Italy South. Alternatively, we can use just two NUTS1 from Italy to represent what you want Italy North and Italy South
> Best regards
> Andrés

Well, same as for the aggregate regions that I added , there is a related definition
Regarding italy I would have
ITN = 52_IT + 53_IT 
ITS =  54_IT+55_IT+56_IT+98_IT
and then I would need to add the correspondence between the ehighway clusters (52_IT...) and the NUTS2 regions (tables are available)
If I would use other separation into north and south then I would lose my data, because I have them available at that granularity.

danielhuppmann:
Thank you @sandrinecharousset and @arght for your comments on additional sub-country regions - however, that discussion should not be part of this PR (which is about variables). Instead, @sandrinecharousset, can you please start a [new issue](https://github.com/openENTRANCE/nomenclature/issues/new) to discuss?

danielhuppmann:
Responding to @erikfilias:

> maximum capacity of a pumped-storage hydropower plant
> `Maximum Capacity|Secondary Energy|Electricity|Hydro|Pumped Storage|Power Plant 1`

Yes, in principle that is how this could be added - but a few considerations and comparison to the IAMC 1.5°C variable list (see [here](https://github.com/danielhuppmann/nomenclature/blob/ab2b23afc9df6d2ef8f6436ae47710d799658aa1/variable/technology/technologies.yaml#L27)), where a similar variable is called.

> Capacity|Electricity|Hydro

1. Having both `Secondary Energy` and `Electricity` is redundant for this type of information (there is no electricity of final energy of a hydro plant), so it can be omitted
2. `Maximum Capacity` is not clear in the context of a pumped-hydro storage plant - is it storage capacity (reservoir), generation capacity, or pumping capacity?
3. For other power plants, "maximum" generation capacity is redundant in the usual terms (that's why we only used `Capacity` in the IAMC 1.5°C context).

So in short, I would suggest to use this more concise version:

- `Capacity|Electricity|Hydro|Pumped Storage|Power Plant 1` for maximum generation capacity
- `Reservoir Capacity|Electricity|Hydro|Pumped Storage|Power Plant 1` for reservoir capacity
- ...

arght:
Daniel,
It is true that we need to differentiate between generation capacity and pumping capacity and between minimum capacity and maximum capacity, all of them power in MW/GW.
Another question is size of the hydro or pumped-storage hydro reservoir that it can be in GWh or hm3 depending on the model. I think that FANSI probably will use hm3 but many other models will use GWh/TWh
Andrés

IngeborgGraabak:
> About the regions definition: from the readme seems we cannot have intermediates between countries and nuts1 regions. in plan4res we use for one country (italy) aggregates which are inbetween (italy separated in ITN and ITS, for noth italy and south italy). Can we give the possibility of having other regionalisations than nuts?

I see your point. But Nuts are clearly defined. In case we are going to use north italy and south italy, we have to define them

danielhuppmann:
Thanks @arght for raising this valid concern about variables where the "normal" unit depends on the model context. Since this is not directly related to either the folder/file structure or the list of variables used in IAMC 1.5°C scenario ensemble, I have started a new issue (#14) for that discussion.

danielhuppmann:
As discussed in the telco earlier today, I am merging this PR because there are no objections to the folder and file structure. Open discussions in this discussion thread have been moved to specific issues.

PR #17, opened 2020-03-31
Adding specifications for sub-annual time resolution

### LINKED ISSUE #5: Define structure for subannual timeslices
This should provide a first set of commonly used subannual timeslices.

danielhuppmann:
This PR adds a subfolder for the sub-annual dimension of the data format, using the disaggregation level of "months" for defining a structure of README and yaml file.

closes #5

### DIFF
--- a/subannual/months.yaml
+++ b/subannual/months.yaml
@@ -1,0 +1,37 @@
+# List of months
+
+January:
+   duration: 31. / 365
+
+February:
+   duration: 28 / 365
+
+March:
+   duration: 31 / 365
+
+April:
+   duration: 30 / 365
+
+May:
+   duration: 31 / 365
+
+June:
+   duration: 30 / 365
+
+July:
+   duration: 30 / 365
+
+August:
+   duration: 31 / 365
+
+September:
+   duration: 30 / 365
+
+October:
+   duration: 31 / 365
+
+November:
+   duration: 30 / 365
+
+December:
+   duration: 31 / 365
--- a/subannual/months.yaml
+++ b/subannual/months.yaml
@@ -1,7 +1,7 @@
 # List of months
 
 January:
-   duration: 31. / 365
+   duration: 31 / 365
 
 February:
    duration: 28 / 365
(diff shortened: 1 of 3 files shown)


subannual/months.yaml
```diff
@@ -0,0 +1,37 @@
+# List of months
+
+January:
+   duration: 31. / 365
```
tburandt: Should make much of a difference for the actual content, but the additional '.' bothers me here

tburandt:
Also, we currently face one problem of using variable representative time-series/hours. 

For example, in the current iteration of GENeSYS-MOD, the user can flexibly choose the temporal resolution when starting a model run. Hereby, a full hourly time-series is reduced by taking each nth hour (for example each 49th). This reduced series is then smoothed and rescaled to keep the characteristics of the underlying full hourly time-series. As a result of the reduction algorithm, a new time-series consisting of consecutive hours, each representing a fraction of the year, is created. 

E.g., choosing each 73th hour for the algorithm results in 120 hours/5 days; choosing each 49th hour or 169th hour are creating resulting time-series with 178 hours or 51 hours, respectively. Therefore, a simple assignment of 'summer-day' or 'summer-hour 1' is not entirely possible. A similar problem can occur when using clustered time-series compared to reduced time-series.

subannual/months.yaml
danielhuppmann: thanks, fixed!

arght:
We do something similar in openTEPES. The model is able to be run (by selecting an option) every 2, 3 or 4 hours but we are still using the same hourly definition like this
2030-01-01T01:00:00+01:00
2030-01-01T02:00:00+01:00
2030-01-01T03:00:00+01:00
2030-01-01T04:00:00+01:00
2030-01-01T05:00:00+01:00
2030-01-01T06:00:00+01:00
and we puick up only the hours corresponding to the "active" ones (every 2, 3 or 4)
Andrés

danielhuppmann:
Thanks @tburandt and @arght for these comments. The simplest approach here would be to have one category for each level of temporal resolution that will actually be used in the project (not necessarily every level that *could* be implemented by each model).

I'd suggest that I merge this PR, as this defines a structure to be followed, and then either of you starts a new PR for the first temporal resolution category that you intend to use in the project - and then we discuss naming conventions etc. in that PR.

danielhuppmann:
Merging as no objections to my earlier comment

PR #18, opened 2020-04-03
add explanation on year-value in the data format

### LINKED ISSUE #9: Improve description of data format for year/value on landing page
Per bilateral communication with @tburandt, the section on **Timeseries data dimensions** is not sufficiently clear in the top-level [README](https://github.com/openENTRANCE/nomenclature/blob/master/README.md) because **year** and **value** are not mentioned. I didn't include them in my first version because they do not require subfolders with more detailed explanation, and I was worried about creating confusion about *long vs. wide table formats*.

So how to add this information in a concise and intuitive way:

1. Add entries to the 
   > 7 Year
   > 8 Value
   
   But this risks confusion because it might be understood that it has to be a long format (it can be either long or wide).

2. Come up with a short description after the list of dimensions (good formulations?)

3. Option 2 but also add example tables (one long, one wide) similar to what is in the [pyam-docs](https://pyam-iamc.readthedocs.io/en/stable/data.html) or [this tutorial](https://pyam-iamc.readthedocs.io/en/latest/tutorials/data_table_formats.html) -but this has the risk of making the top-level README very long...

danielhuppmann:
This PR adds a short explanation of the year-value components in the data format section of the top-level README.

Preview available [here](https://github.com/danielhuppmann/nomenclature/blob/structure/year-value/README.md)

closes #9

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,20 @@ In the data format, every timeseries is described by six dimensions (codes):
 6.	Subannual (optional, default 'Year')<sup>[1]</sup> -
     [more information](subannual)
 
+In addition to these six dimensions, every timeseries is described by
+a set of year-value pairs.
+The resulting table can be either shown in **wide format**
+(see example below, with *years* as columns)
+or in **long format** (two columns *year*  and *value*).
+
+| **model** | **scenario** | **region** | **variable**   | **unit** | **subannual** | **2005** | **2010** | **2015** |
+|-----------|--------------|------------|----------------|----------|---------------|---------:|---------:|---------:|
+| MESSAGE   | CD-LINKS 400 | World      | Primary Energy | EJ/y     | Year          |    462.5 |    500.7 |      ... |
+| ...       | ...          | ...        | ...            | ...      | ...           |      ... |      ... |      ... |
+
+<small>Data via the [IAMC 1.5°C scenario explorer](https://data.ene.iiasa.ac.at/iamc-1.5c-explorer),
+showing a scenario from the [CD-LINKS](https://www.cd-links.org) project.</small>
+
 [1] *The index 'Subannual' is an extension of the original format introduced by
 the openENTRANCE project to accomodate data at a subannual temporal resolution.*
 
--- a/README.md
+++ b/README.md
@@ -55,18 +55,18 @@ In the data format, every timeseries is described by six dimensions (codes):
     [more information](subannual)
 
 In addition to these six dimensions, every timeseries is described by
-a set of year-value pairs.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,6 +54,21 @@ In the data format, every timeseries is described by six dimensions (codes):
 6.	Subannual (optional, default 'Year')<sup>[1]</sup> -
     [more information](subannual)
 
+In addition to these six dimensions, every timeseries is described by
```
erikfilias, line 57: time-series

danielhuppmann: It seems that there is no clear preference between *timeseries*, *time series* and *time-series*, and I don't see much difference in readability (in contrast to sub-annual where I have used the hyphen-version in text but not in the column name). 

For our use case, the term is mostly used as *timeseries data* where *time series data* looks a bit weird (to me at least) - so I have until now tried to always use the one-word spelling (also in the [pyam docs](https://pyam-iamc.readthedocs.io)). So I'd prefer to keep it as is for consistency.

Objections?

danielhuppmann:
> Any date format in the sub-annual column could be used, no? Maybe, We could suggest the use of a standard format such as the ISO 8601 (e.g. 1997-07-16T19: 20: 30 + 01: 00).
> Link: https://www.w3.org/TR/NOTE-datetime

**pyam** can handle continuous-time formats like ISO 8601, but for writing it to the database, we would need to split out the year and the remaining part. There is a function that does part of this, which could be extended (see [IamDataFrame.swap_time_for_year()](https://pyam-iamc.readthedocs.io/en/stable/_modules/pyam/core.html#IamDataFrame.swap_time_for_year)).

README.md
erikfilias: No, It's ok for me.

erikfilias:
Wow, I see. It will help us a lot.
My idea is related to provide a standard format and avoid the use of several date formats. Or, how do you see it?

danielhuppmann:
> My idea is related to provide a standard format and avoid the use of several date formats.

Having one standard format won't work, because some models use representative timeslices that cannot be converted to continuous-time representation (like `summer-day`, `winter-night`). Anyway, this discussion should be moved to a new issue, as it is not directly related to this PR.

danielhuppmann:
merging as approved by @tburandt & @erikfilias

arght:
Daniel,
It depends on what they mean by summer-day and winter-night. If it is a series of hours of a certain day is just a question of selecting those specific hours and, therefore, the stardard definition will work. If it is another time definition then it may not work
Andrés

Issue #6, opened 2020-03-16
Review variables/units and add relevant nomenclature for `plan4EU` model


danielhuppmann:
This will review the first set of variables/units and add a list of variables relevant for the `plan4EU` model. It should not add (at this stage) all required variables (to not have a PR too large to review), but rather see whether the process for adding variables and units is sufficiently simple and intuitive.

depends on #4

sandrinecharousset:
About the Energy/Energy-final variables:
I find a bit ambiguous the way some variables are described. eg: 
Final Energy|Industry|Electricity:
   description: Final energy consumption by the industrial sector of electricity (including on-site solar pv), excluding transmission/distribution losses => I asked myself when reading that 'what is the industrial sector of electricity? => I Believe what is mentioned here is the electricity consumption of the industrial sector? Not the final Energy needs of the electrcity industry sector…. If yes I would change the order of the sentence:
electricity share in the final Energy consumption of the industrial sector or something like that… (not completely sure…) 
The same comment applies to all Final Energy|Industry|*  where it may be ambiguous as there exist industries that produce the final Energy which is described. (and even to others such as Final Energy|residential and Commercial|electricity, because there exist residentials that produce electricity….. but I admit this would require a Strange brain….

Final Energy|Non-Energy Use : the description is "Final energy consumption by the non-combustion processes" : I am not completely sure to understand what that means : does it refer to Something (eg biomass) that is used to produce eg food? (Something not related to Energy)

sandrinecharousset:
@danielhuppmann sorry forgot to mention you in the comment above. Did not use pull request as I am not sure of the change I would require.....

danielhuppmann:
thanks @sandrinecharousset for these comments!

> Final Energy|Industry|Electricity

You are right, this formulation is confusing. It should probably be just changed to "Total electricity consumption of industry"

> Final Energy|Industry

This should include only industry sectors that do not produce energy. The own consumption of the electricity sector, for example, should be covered by the difference between primary-energy input and secondary-energy output.

> Final Energy|Non-Energy Use

I would interpret this as all consumption of fuels that were previously considered primary/secondary energy carriers. So crude oil that is used for plastics - yes. Biomass that is used for food - no.

Hope this helps and is enough to start a PR with suggested changes.

sandrinecharousset:
@danielhuppmann is thank you! Will do the PR once we have clarified the way of using branches and forks........

PR #20, opened 2020-04-08
addition of subcountriy regions & electricity final energy variables

sandrinecharousset:
Added subregions and variables relevant for plan4res
+ proposed new definition of final energy variables that were ambiguous

partly implementing #6


region/README.md
```diff
@@ -75,6 +75,7 @@ One set of disaggregation of countries follows the
 [NUTS 2021 classification](https://ec.europa.eu/eurostat/web/nuts/background)
 used by Eurostat.
 
+<<<<<<< HEAD
```
danielhuppmann, line 78: this is an artefact from a merge conflict and can be removed

```suggestion

```

region/README.md
```diff
@@ -129,8 +130,7 @@ for n3, mapping in nuts3_codelist.items():
 ```
 
 ### Other sub-country area classification
-
-*To be added at a later stage*
+=======
```
danielhuppmann, line 134: This is also an artefact from a merge conflict

```suggestion

```

variable/energy/energy-final.yaml
```diff
@@ -11,87 +12,94 @@ Final Energy|Electricity:
...
 Final Energy|Industry|Hydrogen:
    description: Final energy consumption by the industrial sector of hydrogen
+   # ambiguous: is it industries that produce hydrogen (and thus need energy to produce it) or that consume hydrogen as an energy source
    unit: EJ/yr
 
 Final Energy|Industry|Liquids:
    description: Final energy consumption by the industrial sector of refined
                 liquids (conventional & unconventional oil, biofuels,
                 coal-to-liquids, gas-to-liquids)
+   # ambiguous: is it industries that produce liquids (and thus need energy to produce it) or that consume liquids as an energy source
```
danielhuppmann, line 84: maybe just switching the description to "consumption of refined liquids by the industrial sector" reduces the ambiguity?

region/subcountries.yaml
```diff
@@ -0,0 +1,114 @@
...
+
+# e-highway2050 clusters
+# e-highway2050 is a EU-funded project (2012-2015) whose objective was to provide a 
+# modular and robust expansion plan for the Pan-European Transmission Network from 2020 to 2050
+# In particular it defined a cluster model of the Pan-European transmission grid  (D2.2)
+# See https://docs.entsoe.eu/baltic-conf/bites/www.e-highway2050.eu/e-highway2050/
+
+# e-highway clusters were used in the plan4res project (www.plan4res.eu) to create its public dataset
+# we will describe here only the clusters that are used in plan4res public dataset, which are those
+# describing france, germany and itajy. For the rest of europe, country or regional aggregation is used in plan4res
```
danielhuppmann, line 15: Fixing a typo and capitalising the country names for consistency
```suggestion
# describing France, Germany and Italy. For the rest of europe, country or regional aggregation is used in plan4res
```

sandrinecharousset:
@danielhuppmann Regarding the regions names FR_14.... these are the original region names as they can be found in th ehighway2050 public datasets. Those 'regions' are electrical regions and are not equal to regions that have a name. Ile de france works as FR_23 contains exactly the same NUTS3 as the administrative region Ile de France, so Ile de France=FR_23=FR10 which is the NUTS2 denomination; But this is not the case for all regions, but for other regions it does not fit at all: eg  FR_14 which is the extreme south-west part contains a part of nouvelle-aquitaine and a part of occitanie, then I would have no clue of how to give a name..... The only really accurate stuff that I can do is give the list of NUTS3.

sandrinecharousset:
@danielhuppmann  regarding the variables, I agree with your proposal of using source (energy source sounds betterà, use , sector as attributes (first time I use yaml also....);
Regarding the definition Total final energy consumption by all end-use sectors and all
                 fuels, excluding transmission/distribution losses, I find  ambiguous to write energy consumption by 'list of users' and 'list of fuels' because it makes me think that it is energy used to produce fuel.... (even if in tht case it can be resolved more easily); if we have the attributes below I would here ommit the by all fuels, or write total (all fuels) energy consumption by all end users excluding transmission/distribution losses

danielhuppmann:
@sandrinecharousset responding to the variable discussion
> if we have the attributes below I would here ommit the by all fuels, or write total (all fuels) energy consumption by all end users excluding transmission/distribution losses

Agree to remove "all fuels" in this case mentioned. Ideally, short and concise is better, as long as there is each description is still unique. The additional attributes should not replace a comprehensive description, but only be seen as additional info for clarity and easier workflows, (automatic sorting, etc.)

danielhuppmann:
@sandrinecharousset responding to the region names discussions

Is there any way to come up with somewhat descriptive region names than `24_FR`? If that is not possible, I think even `France|Grid node 24` would be more intuitive - and the mapping to the highways publication can be provided via a mapping.

variable/energy/energy-final.yaml
sandrinecharousset: Yes, agree

sandrinecharousset:
> @sandrinecharousset responding to the region names discussions
> 
> Is there any way to come up with somewhat descriptive region names than `24_FR`? If that is not possible, I think even `France|Grid node 24` would be more intuitive - and the mapping to the highways publication can be provided via a mapping.

@danielhuppmann 
Agree with the general idea. We could call it node or cluster, I have no real preference; 
Regarding the mapping how should it pbe provided?

arght:
Sandrine,
At the beginning we proposed to use these geographical levels: country, NUTS1, NUTS2, NUTS3, nodes, districts, communities, city block, end users in order for all the models to play with geographical information.
I understand that in the case of 24_FR you are proposing a node that can have a longitude and latitude. For nodes, any name would be valid
Andrés

sandrinecharousset:
> Sandrine,
> At the beginning we proposed to use these geographical levels: country, NUTS1, NUTS2, NUTS3, nodes, districts, communities, city block, end users in order for all the models to play with geographical information.
> I understand that in the case of 24_FR you are proposing a node that can have a longitude and latitude. For nodes, any name would be valid
> Andrés
@arght NUTS1, 2 , 3 are already included in the nomenclature. Here the question s of adding other subcountry regions. All those that I want to include (because I need them), can be defined as an aggregation of NUTS3 regions. Example 24_FR=FRF22 + FRF24 + FRC11 + FRC13 + FRC14, where all those are NUTS3 regions. The current discussion is about the use of the name 24_FR which is not very intuituve (NUTS names are not intuitive neither but they exist and you can easilky find maps). It is quite difficult to map those 24_FR like regions to already existing regions somewhere, and I also do not have the coordinates (as it is regions and not physical nodes, it is wide areas, so coordinates are a bit tricky to use....), but the link to NUTS completely determines what that region is.
Daniel proposed that instead of using the names from ehighway20050 (which are not normalised, and even in the project documents, they use 2 different namings for the same regions), we use something more 'human readable', including the informaiton we have (we know FR_24 is in France) ... I understand that you would be ok with that naming (France| Grid Node FR24)
Last think, if we use Grid node 24, I would think you have node 1, 2, ..., 24,... but this is absolutely ot the case....and I could not find the logic there.....

arght:
Then, I guess there are two possible solutions:
One it is to create somehting in between NUTS2 and NUTS3 (called subcountry regions)
The other will be to make a relocation of these regions into a particular NUTS3 that approximately corresponds to it (e.g., change 24_FR to FRC11 everywhere)
Will any of these possibilities sounds fine for you?
andrés

sandrinecharousset:
> Then, I guess there are two possible solutions:
> One it is to create somehting in between NUTS2 and NUTS3 (called subcountry regions)
> The other will be to make a relocation of these regions into a particular NUTS3 that approximately corresponds to it (e.g., change 24_FR to FRC11 everywhere)
> Will any of these possibilities sounds fine for you?
> andrés

@arght I do not understand what you want to do. We already have a solution which consists in adding those regions as they are (if we change them all the dataset will become unusable). The idea is to allow each modelling team to add the regions they need. The dataset that i have and want to use as a complement to the scenarios (because not everything that I need will be available in WP3) exist at the granularity of those regions, with lots of data. 
Changing FR-24 to FRC11 is to my mind the thing not to do because FR24 is much bigger than FRC11, has much more demand and installed capacity then if I would change that, imagine that I find some data that describe the genmix at the scoope of NUTS3.... I would have a big problem...
The only thing we wanted to discuss here is finding appropriate names for that kind of regions, not replacing them by anything else :-)

sandrinecharousset:
> Then, I guess there are two possible solutions:
> One it is to create somehting in between NUTS2 and NUTS3 (called subcountry regions)
> The other will be to make a relocation of these regions into a particular NUTS3 that approximately corresponds to it (e.g., change 24_FR to FRC11 everywhere)
> Will any of these possibilities sounds fine for you?
> andrés

@arght sorry answered too fast. Your first solution is exactly what we are doing, with the difference that we are not creating sth new as it already exists. We are implementing it in our nomenclature. THe idea being that if other kinds of regions are needed that are not available, we can add them, provided that the have a non ambiguous description (which is the case here because it relies on nuts3)

arght:
OK
so, then the solution would be to add the subcountry regions
In such a case I think that France|Subregion 24 would be a possibility
Andrés

sandrinecharousset:
Exactly
In the name I liked the dea of having the workd grid. The reason is that those regions are artificial and have been constructed by TSOs only looking at alectricity flows. They are 'electricity grid regions' that mean something only from the point of view of electricity grids....

sandrinecharousset:
SubGrid?

arght:
subcountry, I am sorry

arght:
subgrid, it is also fine

arght:
are these regions homogeneous from the point of view of voltages or marginal prices, for example?

sandrinecharousset:
Only transmission grid is included , the idea is to build an equivalent grid with less nodes. In the results you have the min/max flows and impédances of equivalent lines
I can send you the document where the method to build them is described. Was a work by all European tsos 
Nothing to do with prices

arght:
OK
Thanks
It can be useful

danielhuppmann:
responding to @sandrinecharousset and the region naming:

> Agree with the general idea. We could call it node or cluster, I have no real preference;
Regarding the mapping how should it be provided?

I would suggest the following structure (extending the example I gave above (the text after `#` are comments and should not be implemented):
```
France|Grid node 23  # use a hierarchical structure (France|...) only if the node name is not intuitive
   country: France  # add country explicitly to allow easy mapping and filtering
   definition: ...  # using a list of any other region names already defined in any yaml file (NUTS-x, ...)
   note: This node was defined and used in the ehighway2050 public datasets
   ehighways: 23_FR
```

sandrinecharousset:
> responding to @sandrinecharousset and the region naming:
> 
> > Agree with the general idea. We could call it node or cluster, I have no real preference;
> > Regarding the mapping how should it be provided?
> 
> I would suggest the following structure (extending the example I gave above (the text after `#` are comments and should not be implemented):
> 
> ```
> France|Grid node 23  # use a hierarchical structure (France|...) only if the node name is not intuitive
>    country: France  # add country explicitly to allow easy mapping and filtering
>    definition: ...  # using a list of any other region names already defined in any yaml file (NUTS-x, ...)
>    note: This node was defined and used in the ehighway2050 public datasets
>    ehighways: 23_FR
> ```

THanks @danielhuppmann 
I agree with the concept and will implement it
Last thing, we discussed with @arght  about the name of those regions and it appears that using 'node' may create some misunderstandings. We suggested SubGrid. What do you think?
It would be;
France|SubGrid 23  
>    definition: ...  # using a list of any other region names already defined in any yaml file (NUTS-x, ...)
>    note: This node was defined and used in the ehighway2050 public datasets
>    ehighways: 23_FR

danielhuppmann:
> We suggested SubGrid. What do you think?

@sandrinecharousset, sounds ok to me - but I suggest to follow the [variable naming conventions](https://github.com/openENTRANCE/nomenclature/tree/master/variable#naming-convention-for-variables) for consistency and either write `Subgrid` (no camel-case) or `Sub-Grid` (separate words capitalized and connected by a hyphen).

sandrinecharousset:
> > We suggested SubGrid. What do you think?
> 
> @sandrinecharousset, sounds ok to me - but I suggest to follow the [variable naming conventions](https://github.com/openENTRANCE/nomenclature/tree/master/variable#naming-convention-for-variables) for consistency and either write `Subgrid` (no camel-case) or `Sub-Grid` (separate words capitalized and connected by a hyphen).

@danielhuppmann yes of course, sorry for forgetting that convention :-)

PR #24, opened 2020-04-17
Addition of variables for plan4EU

### LINKED ISSUE #6: Review variables/units and add relevant nomenclature for `plan4EU` model

sandrinecharousset:
THis pull request closes #6 
Most variables for plan4EU have been added 
- in energy
- in technology


variable/energy/energy-final.yaml
```diff
@@ -273,3 +273,38 @@ Final Energy|Transportation|Liquids|Oil:
    description: Final energy consumption by the transportation sector of liquid
                 oil products (from conventional & unconventional oil)
    unit: EJ/yr
+
+Final Energy|Electricity|Profile:
```
danielhuppmann, line 277: I'm wondering why this needs an own variable? If the variable `Final Energy` is given for a full year, it is the total consumption over the year - if it is given with an hourly disaggregation, it already fulfils the role of this "profile". So I see this as a duplication.

danielhuppmann, line 277: I'm wondering why this needs an own variable? If the variable `Final Energy` is given for a full year, it is the total consumption over the year - if it is given with an hourly disaggregation, it already fulfils the role of this "profile". So I see this as a duplication.

variable/technology/electricity grid.yaml
```diff
@@ -0,0 +1,25 @@
+# List of variables related to the electricity grid
```
danielhuppmann, line 1: having file names with spaces can create all kinds of problems, so I suggest to rename to `electricity-grid.yaml`

variable/technology/electricity grid.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+Frequency Containment Reserve Demand|Electricity|Grid:
```
danielhuppmann, line 3: This is a very long top-level category name. Not sure what the best approach here is - what do you think of `Grid|Electricity|Demand|Reserve|Frequency Containment` (or may be `Network` for generality?). This will mean that all related variables will be shown next to each other in the Scenario Explorer (where the variable selection is organised in a tree structure).

This suggestion also applies to the other variables proposed here...

variable/technology/technologies.yaml
```diff
@@ -28,7 +76,21 @@ Capacity|Electricity|Hydro:
...
    unit: GW
+   
+Capacity|Electricity|Hydro|Reservoir:
+   description: Total installed electricity generation capacity of hydro power plants with 
+                reservoirs with seasonal storage potential
+   unit: GW
 
+Capacity|Electricity|Hydro|Pumped Storage:
+   description: Total installed electricity generation capacity of hydro power plants with 
+                pumped storage
```
danielhuppmann, line 87: suggest to rewrite for clarity

```suggestion
   description: Total installed electricity generation capacity of pumped-storage hydro power plants
```

variable/technology/technologies.yaml
```diff
@@ -28,7 +76,21 @@ Capacity|Electricity|Hydro:
...
+   unit: GW
+   
+Capacity|Electricity|Hydro|Run of River:
```
danielhuppmann, line 90: I can't say why, but this sounds more "correct" to me...

```suggestion
Capacity|Electricity|Hydro|Run-of-River:
```

variable/technology/technologies.yaml
```diff
@@ -460,3 +522,570 @@ Investment|Energy Supply|Liquids|Oil:
                 capturing equipment should be included but not the one on
                 transport and storage.
    unit: billion US$2010/yr
+
+# description of thermal power plants 
+
+Number Units|Electricity|Biomass|Traditionnal:
```
danielhuppmann, line 528: I'm wondering whether `Number of Units` would be more readable...

fixing a typo
```suggestion
Number Units|Electricity|Biomass|Traditional:
```

variable/energy/energy-final.yaml
sandrinecharousset: @danielhuppmann I made a mistake here with the unit. This is not MWh but percentages and this variable is kind of a loadfactor (if you multiply hourly values by yearly MWh you get hourly MWh). The idea is that those profiles are "fixed" and they include all the seasonal, weekly, daily variations as well as the stochasticity (they are scenarised). When performing a case study, the real thing that is put into the plan4EU core is profile multiplied by yearly energy. Goig to correct that.

danielhuppmann: Thanks for the clarification @sandrinecharousset! If this is a relative value, please clarify as a `note` attribute what the 100%-value is (average load, peak load).

sandrinecharousset: @danielhuppmann I have just commited a change here

variable/technology/electricity grid.yaml
sandrinecharousset: @danielhuppmann done

variable/technology/electricity grid.yaml
sandrinecharousset: Agree, done

variable/technology/technologies.yaml
sandrinecharousset: agree

arght:
Sandrine,
Is there a definition of pumped-storage hydro as technology, CAES or in general ESS?

sandrinecharousset:
@danielhuppmann help!, a conflict appeared in energy-final which I do not understand....

sandrinecharousset:
@arght regarding the hydro, in our dataset we have 3 kinds of hydro
- Hydro reservoir is seasonal storage ; this is an aggregation of all seasonal reservoirs per country (or aggregated countries). Some of them also have pumpping capacity. 
- pumped-storage which is hydro that has some storage (and pumpin capacity)  but a smaller time-scale (weekly or daily). We do not refer to technologies here. 
- Run-of-river without storage capacity
It is very important to differentiate those 3 kinds becaus they are not managed the same way (both from an operation point of view and from a mathematics point of view):
- hydro reservoir are used for seasonal transfers of energy (or in case of very large reservoirs, multi-year transfers); in our model they are those that have bellman values
- pumped storage is used for short term energy transfer (weekly, daily). We do not compute bellman values
- run of river is not used for transfer of energy

danielhuppmann:
> @danielhuppmann help!, a conflict appeared in energy-final which I do not understand....

Fixed it for you!

For an explanation what happened: your branches `refs/subcountry` and `defs/plan4eu-variables` started from the same point in the history and both made changes to a particular line in the same file, so git did not know how to get these two changes together and alerted you that this branch cannot be merged into `master`(which already has the changes from `refs/sbcountry`).

I used the GitHub "Resolve conflict" function to tell git how to correctly merge the changes. Looking at it in a visual user interface (e.g. GitHub desktop, GitKraken) can help to illustrate the problem of the timeline and the solution.

sandrinecharousset:
> > @danielhuppmann help!, a conflict appeared in energy-final which I do not understand....
> 
> Fixed it for you!
> 
> For an explanation what happened: your branches `refs/subcountry` and `defs/plan4eu-variables` started from the same point in the history and both made changes to a particular line in the same file, so git did not know how to get these two changes together and alerted you that this branch cannot be merged into `master`(which already has the changes from `refs/sbcountry`).
> 
> I used the GitHub "Resolve conflict" function to tell git how to correctly merge the changes. Looking at it in a visual user interface (e.g. GitHub desktop, GitKraken) can help to illustrate the problem of the timeline and the solution.

THanks @danielhuppmann , I was feared of again messing everything.... I had a look at gitkraken and understood what happened

variable/energy/energy-final.yaml
sandrinecharousset: @danielhuppmann I changed it; added a note attribute as suggested, and regarding the unit, I think that is is not either a percentage as 100% does not have a meaning. Those time series are time series of coefficients which, when multiplied by a yearly energy in MWh, give time series of MWh. I wrote that in the note.

Issue #14, opened 2020-03-26
Dealing with variables related to units in different dimensions


danielhuppmann:
This issue extracts one discussion from the review of #12 (see [this comment](https://github.com/openENTRANCE/nomenclature/pull/12#issuecomment-604016277)), which needs to have its own dedicated discussion.

There are variables that may be measured in units in different dimensions, the first example being the storage capacity of a reservoir. This can be measured as volume of water (relevant for hydrological models) or as energy stored (relevant for electricity system models). In these cases, the values can be converted in principle but this requires additional information and depends on the reservoir and turbine characteristics.

[For clarification, this is different from unit conversion from exajoule to megawatthours, which is a straightforward multiplication]

Two options to deal with this:
1. use the same variable e.g. `Reservoir Capacity|Electricity|Hydro|Pumped Storage|Power Plant 1` but with different units depending on the model context
    - not having the same (type of) unit for a specific variable makes it difficult/impossible to do automatic unit conversion
2. use distinct variables connected to different units
    - in this case, we could use `Storage Capacity` for energy and `Reservoir Capacity` for water volume
    - but this might become complicated if there are many such cases

@arght @sandrinecharousset @erikfilias @tburandt 

**Question 1** - do you have other examples of such conflicts?
**Question 2** - any preferences for option 1 or 2?

arght:
I prefer option 2 because, to my understanding, there is no direct connection between both variables: reservoir capacity and storage capacity. In the first case, the model is dealing with volume units to represent the reservoir management (which will be the case of a hydro scheduling model) and, in the second case, not such detail is considered (because the focus of the model is not on hydro scheduling, e.g., openTEPES or because the model works with aggregated/equivalent reservoirs)

sandrinecharousset:
We (EDF) usually use the name Volume when it is a real volume, of water, measured in hm3, and storage when it is about energy. Direct conversion is always difficult if not impossible:
- when it is a real reservoir, the efficiency of the turbine depends of the amount of water left in the reservoir; it is even more complex when dealing with complex cascading systems
- same as @arght  reservoirs used in models are not always real - eg often we model national reservoirs, here the conversion requires a whole model
For me the volume of water in a reservoir is somtthing that is different of the amount of energy that this reservoir is storing, then would deserve different names. Moreover, the data that are using one of the units (eg the volumes in hm3) will never be sed by the models that require storages in GWh, which is another justification for having different names

erikfilias:
[...]

PR #28, opened 2020-05-11
Set up tests and continuous integration

danielhuppmann:
This PR adds a first test to ensure that all yaml files can be parsed correctly and sets up a GitHub actions workflow to execute the test on any branch or PR.

The PR also fixes those issues already introduced in the nomenclature to date.


danielhuppmann:
Thanks @erikfilias for this comment. The change proposed in this PR is just for correct yaml-syntax - `%` is a reserved character so using it in the previous way (without quotes) will not allow for the files to be read correctly.

We already had some discussions about allowing several units in #14, and we should move the discussion if/how to allow this in that issue.

PR #30, opened 2020-05-15
Included new consumption types

paopis:
Included consumption of electricity in monetary value
Included general consumption and consumption of electricity as percentage changes relative to 2020.

### DIFF
--- a/variable/economy/economy.yaml
+++ b/variable/economy/economy.yaml
@@ -4,6 +4,18 @@ Consumption:
    description: Total consumption of all goods, by all consumers in a region
    unit: billion US$2010/yr
 
+Consumption|Electricity:
+   description: Total consumption of electricity, by all consumers in a region
+   unit: billion US$2010/yr
+   
+Consumption|Percentage change relative to 2020:
+   description: Total consumption of of all goods, by all consumers in a region
+   unit: '%'
+
+Consumption|Percentage change relative to 2020|Electricity:
+   description: Total consumption of electricity, by all consumers in a region
+   unit: '%'
+
 Discount rate|Buildings:
    description: Discount rate for investments into residential and commercial
                 buildings


variable/economy/economy.yaml
```diff
@@ -4,6 +4,18 @@ Consumption:
    description: Total consumption of all goods, by all consumers in a region
    unit: billion US$2010/yr
 
+Consumption|Electricity:
```
danielhuppmann, line 7: How is that different from `Final Energy|Electricity`? Or shouldn't this be expenditure rather than consumption?

variable/economy/economy.yaml
```diff
@@ -4,6 +4,18 @@ Consumption:
...
+   description: Total consumption of electricity, by all consumers in a region
+   unit: billion US$2010/yr
+   
+Consumption|Percentage change relative to 2020:
```
danielhuppmann, line 11: The percentage change is clearly a derived measure, so this variable is duplicating information - it would be more rigorous to only report absolute quantities and let users compute relate changes themselves.

paopis: I only wanted to add a measure of the change of physical quantities, but it does not make sense when we consider the aggregated consumption. Therefore I agree with your comment and leave it as a monetary flow

variable/economy/economy.yaml
paopis: I cannot seem to locate the "Final Energy|Electricity" variable in the economy.yaml file.
Am I looking at some wrong version?

danielhuppmann: It's in the file `energy/energy-final.yaml`

danielhuppmann:
ping @paopis, will you continue working on this PR or can it be closed (without merging)?

paopis:
Sure. I can use variables available in the energy-final.yaml. The PR can be closed. Thanks!

PR #31, opened 2020-05-18
Add demand side flexibility potentials

omnipotent12:
For case study 1 we will need variables denoting the potential and related costs of demand side flexibility. Lines 1093 - 1118 contain a first take on how such variables could be specified. @sandrinecharousset, @jed

### DIFF
--- a/variable/technology/technologies.yaml
+++ b/variable/technology/technologies.yaml
@@ -1095,4 +1095,4 @@ Rate Automatic Frequency Restoration Reserve|Electricity|Wind|OffShore:
 Load Shifting Potential|Electricity|Residential|AC:
    description: Maximum shiftable load in a given region/time frame,
                 due to households changing their air conditioning usage.
-   unit: 'MW'
+   unit: MW
--- a/variable/technology/technologies.yaml
+++ b/variable/technology/technologies.yaml
@@ -1089,3 +1089,30 @@ Rate Automatic Frequency Restoration Reserve|Electricity|Wind|OffShore:
    description: Percentage of the Maximum Active Power that can be devoted to Automatic Frequency Restoration Reserve
                 for a Wind OffShore power plant
    unit: '%'
+   
+# description of Demand flexibility potentials
+Number of Units|Electricity|Residential|Demand Flexibility:
+   description: Number of units of kind "simple Hydro reservoir "
+   unit: 
+   
+Maximum Active power|Electricity|Residential|Demand Flexibility:
+   description: Potential for demand response in a given region/time frame, 
+                expressed as % decrease in residential electricity demand
+   unit: %
+   
+Maximum Storage|Electricity|Residential|Demand Flexibility:
+   description: Potential for demand dispatch in a given region/time frame,
+                expressed as % increase in residential electricity demand 
+   unit: %
+   
+Variable Cost|Electricity|Residential|Demand Flexibility:
+   description: Marginal cost of Demand Flexibility in residential electricity demand
+   unit: €/% change
+   
+Capital Cost|Electricity|Residential|Demand Flexibility|Maximum Active power:
+   description: Fixed costs associated with demand flexibility potentials (e.g. installing IoT devices)
+   unit: €/%
+   
+Capital Cost|Electricity|Residential|Demand Flexibility|Maximum Storage:
+   description: Fixed costs associated with demand flexibility potentials (e.g. installing IoT devices)
+   unit: €/%
--- a/variable/technology/technologies.yaml
+++ b/variable/technology/technologies.yaml
@@ -1092,7 +1092,7 @@ Rate Automatic Frequency Restoration Reserve|Electricity|Wind|OffShore:
    
 # description of Demand flexibility potentials
    
-Load Shifting Potential|Electricity|Residential|AC:
+Load Shifting Potential|Electricity|Residential|Air Conditioning:
    description: Maximum shiftable load in a given region/time frame,
                 due to households changing their air conditioning usage.
    unit: MW
--- a/variable/technology/technologies.yaml
+++ b/variable/technology/technologies.yaml
@@ -1091,28 +1091,27 @@ Rate Automatic Frequency Restoration Reserve|Electricity|Wind|OffShore:
    unit: '%'
    
 # description of Demand flexibility potentials
-Number of Units|Electricity|Residential|Demand Flexibility:
-   description: Number of units of kind "simple Hydro reservoir "
-   unit: 
    
 Maximum Active power|Electricity|Residential|Demand Flexibility:
    description: Potential for demand response in a given region/time frame, 
-                expressed as % decrease in residential electricity demand
-   unit: %
+                expressed as '%' decrease in residential electricity demand,
+                analogous to inputting power from a storage unit.
+   unit: '%'
    
 Maximum Storage|Electricity|Residential|Demand Flexibility:
    description: Potential for demand dispatch in a given region/time frame,
-                expressed as % increase in residential electricity demand 
-   unit: %
+                expressed as '%' increase in residential electricity demand,
+                analogous to drawing power into storage unit.
+   unit: '%'
    
 Variable Cost|Electricity|Residential|Demand Flexibility:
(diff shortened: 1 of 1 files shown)


jed:
LGTM, 👍.

variable/technology/technologies.yaml
```diff
@@ -1089,3 +1089,30 @@ Rate Automatic Frequency Restoration Reserve|Electricity|Wind|OffShore:
...
+   
+# description of Demand flexibility potentials
+Number of Units|Electricity|Residential|Demand Flexibility:
+   description: Number of units of kind "simple Hydro reservoir "
+   unit: 
+   
+Maximum Active power|Electricity|Residential|Demand Flexibility:
+   description: Potential for demand response in a given region/time frame, 
+                expressed as % decrease in residential electricity demand
+   unit: %
```
danielhuppmann, line 1101: the %-character needs to be put in between quotes (see line 1091 above) for yaml files to be parsed correctly (as indicated by the failing tests)

variable/technology/technologies.yaml
```diff
@@ -1089,3 +1089,30 @@ Rate Automatic Frequency Restoration Reserve|Electricity|Wind|OffShore:
...
+   
+Maximum Storage|Electricity|Residential|Demand Flexibility:
```
danielhuppmann, line 1103: I'm confused between the variable name "Maximum storage" and the description "Potential for demand dispatch"...

omnipotent12: Following Gils 2014 (https://doi.org/10.1016/j.energy.2014.02.019) we can conceptualize demand flexibility as an electricity storage unit. Maximium storage is akin to a storage unit pulling power from the grid, i.e. increasing demand at that point in time or 'dispatching demand'. I am trying to use the already defined higher level variables' nomenclature here.

Does that work, or would another conceptualization be preferred?

danielhuppmann: It makes sense when you interpret flexibility as temporal shifting. Still, it would be good to clarify the description (clearly stating that this flexibility is interpreted as storage), and/or add a `note` or `seealso` attribute with the scientific reference (this is also the attribute used as [sphinx markup directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-seealso))

variable/technology/technologies.yaml
```diff
@@ -1089,3 +1089,29 @@ Rate Automatic Frequency Restoration Reserve|Electricity|Wind|OffShore:
    description: Percentage of the Maximum Active Power that can be devoted to Automatic Frequency Restoration Reserve
                 for a Wind OffShore power plant
    unit: '%'
+   
+# description of Demand flexibility potentials
+   
+Maximum Active power|Electricity|Residential|Demand Flexibility:
```
sandrinecharousset, line 1095: If the variable is called Maximum Active Power then it should have as unit a power (here MW), and not be a percentage; If we stich to a percentage then the name of the variable should include "rate' or sth else, or I fear it may be confusing with al other "active power" in the nomenclature

variable/technology/technologies.yaml
sandrinecharousset: I think the name should include 'shifting' instead of storage.. Modelling those as storages is a modelling concept but ther could be others. Regarding the demand response, we usually define 2 kinds: 
1/ load curtailement => this consists in "erasing" a certain part of the demand, which may be replaced by another source of local energy (eg local fuel generators which are used to compensate load curtailement which is used mainly for dealing with transmission problems), or not replaced at all (eg for individuals, they agree not to consume if asked to but in exchange get a more interesting price)
2/ Load shifting => this consists in 'moving' a certain amount of energy from one timeframe to another. Typical example for individuals is tariffs that incent them to run the washing machine at night and not during the peaks
I think this kind of vocabulary should be used

variable/technology/technologies.yaml
```diff
@@ -1089,3 +1089,29 @@ Rate Automatic Frequency Restoration Reserve|Electricity|Wind|OffShore:
...
+   description: Potential for demand dispatch in a given region/time frame,
+                expressed as '%' increase in residential electricity demand,
+                analogous to drawing power into storage unit.
+   unit: '%'
+   
+Variable Cost|Electricity|Residential|Demand Flexibility:
+   description: Marginal cost of Demand Flexibility in residential electricity demand
+   unit: €/'%' change
+   
+Capital Cost|Electricity|Residential|Demand Flexibility|Maximum Active power:
```
sandrinecharousset, line 1111: I do not understand why this variable contains both cost and power?

variable/technology/technologies.yaml
sandrinecharousset: Then we may describe the variables in the same spirit than for power plants : we have different kinds of power plants (eg coal, gas...) with the same set of characteristics (maximum power, reserve capabilities, costs....); Here we have different kinds of demand-response which could be load shifting / load curtailment; could apply to different segments of customers (segments are already defined in "energy" variables / and have specific characteristics : costs, amount of power that is curtailed/shifted, amount of energy that can be curtailed/shifted (in that case it is close to a storage)...

omnipotent12:
Hi @sandrinecharousset, thanks for taking a look at that. Your suggestion to represent demand flexibility as a type of storage is exactly what I wanted to do. Perhaps the cleanest way to implement this would be to copy the variables that represent load shifting in plan4EU, as they are implemented in the OpenEntrance nomenclature, and just add corresponding cost parameters (something we hope to generate in CS1), and perhaps modify them a bit. Have loadshifting variables for plan4EU already been added to the nomenclature? I didn't find any in looking through the files.

sandrinecharousset:
> Hi @sandrinecharousset, thanks for taking a look at that. Your suggestion to represent demand flexibility as a type of storage is exactly what I wanted to do. Perhaps the cleanest way to implement this would be to copy the variables that represent load shifting in plan4EU, as they are implemented in the OpenEntrance nomenclature, and just add corresponding cost parameters (something we hope to generate in CS1), and perhaps modify them a bit. Have loadshifting variables for plan4EU already been added to the nomenclature? I didn't find any in looking through the files.

@omnipotent12 no we did not include those specific variables yet. I will try to propose something in the coming days (hopefully on friday); The only purpose is to find relevant names that make people understand this is demand-response and not physical storage, meanwhile keeping very close to what exists regarding seasonal and short term storage. At the end in the core of the model, it will be treated the same way (meaning same way for load curtailement and seasonal hydro storage; and same way for load shifting and short-term storage, which already groups together batteries and short term hydro storage, and should also include EV)

omnipotent12:
@sandrinecharousset, sounds good. I will look out for that update and try to adapte the nomenclature you propose for the purposes of Case study 1 - which focuses on residential demand flexibility.

variable/technology/technologies.yaml
```diff
@@ -1089,3 +1089,10 @@ Rate Automatic Frequency Restoration Reserve|Electricity|Wind|OffShore:
    description: Percentage of the Maximum Active Power that can be devoted to Automatic Frequency Restoration Reserve
                 for a Wind OffShore power plant
    unit: '%'
+   
+# description of Demand flexibility potentials
+   
+Load Shifting Potential|Electricity|Residential|AC:
```
danielhuppmann, line 1095: I'd suggest to spell it out to avoid confusion with "Alternate Current"
```suggestion
Load Shifting Potential|Electricity|Residential|Air Conditioning:
```

variable/technology/technologies.yaml
```diff
@@ -1089,3 +1089,10 @@ Rate Automatic Frequency Restoration Reserve|Electricity|Wind|OffShore:
...
+   description: Maximum shiftable load in a given region/time frame,
+                due to households changing their air conditioning usage.
+   unit: 'MW'
```
danielhuppmann, line 1098: No need to put strings in parentheses... (only special characters like %)
```suggestion
   unit: MW
```

omnipotent12:
Thanks @danielhuppmann. To try to ensure consistency I have made a new PR #51 as a fork from the current master repo. 

Sorry if this is not the right process - still getting used to github.

PR #43, opened 2020-06-05
Addition of output variables for plan4eu

sandrinecharousset:
This pull request includes addition of a new file dealing with the outputs of plan4EU in terms of electricity system operation. This does not include outputs in termes of investment decisions. it includes:
* Commitment of all technologies and demand-side systems (aggregated per technology)
      - Active Power (ie generation schedules)
      - Reserves commitments (Frequency restoration reserve and automatic frequency containment reserve)
      - Inertia available for the system
* Levels of aggregated storages
* Marginal costs of following constraints:
      - demand 
      - reserve requirements (Frequency restoration reserve and automatic frequency containment reserve)
      - inertia requirement
      - max/min flows on lines

### DIFF
--- a/nomenclature/definitions/variable/technology/electrity-operation.yaml
+++ b/nomenclature/definitions/variable/technology/electrity-operation.yaml
@@ -78,6 +78,25 @@ Active power|Electricity|Energy Storage System:
    description: Active Power of energy storage systems, excl. hydro storage
    unit: MWh
    
+# Storage per technology
+# (generally hourly, daily or weekly) level of storage available per technology
+
+Storage|Electricity|Hydro|Reservoir:
+   description: Available storage of "simple Hydro reservoir " units
+   unit: MWh
+   
+Storage|Electricity|Hydro|Pumped Storage:
+   description: Available storage of "Pumped Storage " units
+   unit: MWh
+     
+Storage|Electricity|Energy Storage System:
+   description: Available storage of energy storage systems, excl. hydro storage
+   unit: MWh
+
+Storage|Electricity|Load Curtailement:
+   description: remaining amount of curtailement
+   unit: MWh
+   
 # Operation Cost per technologies
 # i.e.  operation cost of aggregated committments of plants, storages and demand-side systems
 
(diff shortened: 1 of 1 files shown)


nomenclature/definitions/variable/technology/electrity-operation.yaml
```diff
@@ -0,0 +1,460 @@
...
+#   - Demand in ACtive Power
+#   - Demand in reserves
+#   - Demand in inertia
+#   - Flows in transmission lines
+# Demand not served
+
+# Active Power per technologies
+# i.e.  aggregated committed (in general hourly) schedules of plants
+
+Active Power|Electricity|Biomass:
```
danielhuppmann, line 18: is "Active power" different from "Generation"? if no, I'm wondering if using the more intuitive term isn't better...

nomenclature/definitions/variable/technology/electrity-operation.yaml
```diff
@@ -0,0 +1,460 @@
...
+   
+Operation Cost|Electricity|Electricity Storage:
+   description: Operation Cost of energy storage systems, excl. hydro storage
+   unit: €
+
+# Automatic Frequency Restoration Reserve committed per technologies
+# i.e.  Automatic Frequency Restoration Reserve of aggregated committments of plants, storages and demand-side systems
+
+Automatic Frequency Restoration Reserve|Electricity|Biomass:
+   description: Automatic Frequency Restoration Reserve available for  Traditionnal Biomass power plants
```
danielhuppmann, line 210: Is this the amount of AFRR available (could be used, i.e., maximum), or the amount actually contracted? And how will this relate to the amount of reserve required? That would be "Demand|..."?

nomenclature/definitions/variable/technology/electrity-operation.yaml
sandrinecharousset: We hesitated a lot.... The idea is that if Comillas wants further to include reactive power (as was mentioned in a previous discussion), would be more clear with Active Power. For plan4eu, we don't care
@erikfilias what do you think?

nomenclature/definitions/variable/technology/electrity-operation.yaml
sandrinecharousset: This is the amount committed, ie the maximum that can really be used if there is a problem on the grid. It is the result of the optimization (the optimisation decides from the Maximum Power of each plant which part to commit as active power, which part as aFRR, ....); It is different from the contracted amount, but the sum of all committed aFFR on a specific perimeter should be greater than the amount contracted.

nomenclature/definitions/variable/technology/electrity-operation.yaml
```diff
@@ -15,66 +15,90 @@
...
+Active Power|Electricity|Wind|OnShore:
    description: Active power of  Wind Onshore power plants
    unit: MWh
    
-Active power|Electricity|Wind|OffShore:
+Active Power|Electricity|Wind|OffShore:
    description: Active power of  Wind OffShore power plants
    unit: MWh
    
-Active power|Electricity|Load Curtailement:
```
omnipotent12: should be "Curtailment"

sandrinecharousset: Yes, thanks, will change it

nomenclature/definitions/variable/technology/electrity-operation.yaml
```diff
@@ -15,66 +15,90 @@
...
+Active Power|Electricity|Load Curtailement:
    description: Energy curtailed - from the system point of view is equivalent to generation of Active Power
    unit: MWh
    
-Active power|Electricity|Load Shifting:
+Active Power|Electricity|Load Shifting:
```
omnipotent12, line 97: Are both curtailment and load shifting seen as  "eqivalent to generation of active power"? Does this mean that load shifting can take negative values when demand is increased?

As an aside: I think this fits well with what I now proposed for demand flexibility potentials, namely:
Load Shifting Potential|Electricity|Residential|AC

sandrinecharousset: Yes it is, can take negative values. For load shifting if the model decides to move load from instant t1 to instant t2 then  what happens is that -the electricity demand variable does not change  -load shifting(t1) is positive (which is equivalent to decreasing the elec load) - load shifting in t2 is negative, which is equivalent to increasing the elec load
Nadia is working on defining additionnal input variable for feeding the model with also the related constraints (eg time horizon constraints for the shifting....)

omnipotent12: Great, thanks!
looks good to me for a merge with Master.

sandrinecharousset:
> Sandrine,
> Then it should be understood that these are results from a model.
> Shouldn't be the units of aFRR, mFRR and RR in MW or MWh/h?
> Andrés

@arght usually the unit is MWh , and this output is hourly profiles 
and yes this is the result of a model (it is in a file called electricity-operation which only contains 'outputs' of model); In the description I wrote 'considering the commitment decisions taken by the odel', then thought it was clear it is an output?

danielhuppmann:
merging as is and then implement a clean-up in a separate PR later

PR #37, opened 2020-05-28
Adding variables from GENeSYS-MOD power sector

tburandt:
First, I added different types of methane used in electricity production (biomethane, natural gas, and synthetic methane). Therefore, `Secondary Energy|Electricity|Gas` is now an aggregate of all methane/natural gas-type fuels.

Secondly, I found an error in the `power-plant.yaml` file, where coal plants were falsely stated as `|Gas|Coal` and CCGTs were missing. The later has been added as well.

Lastly, I added hydrogen power generation plants in `technologies.yaml`

PR #48, opened 2020-06-09
Adding variables from GENeSYS-MOD power sector

tburandt:
First, I added different types of methane used in electricity production (biomethane, natural gas, and synthetic methane). Therefore, `Secondary Energy|Electricity|Gas` is now an aggregate of all methane/natural gas-type fuels.

I added hydrogen power generation plants in `technologies.yaml`

Lastly, comments from #37 were also included.


danielhuppmann:
@erikfilias @sandrinecharousset can you take a look if this makes sense for you?

sandrinecharousset:
> @erikfilias @sandrinecharousset can you take a look if this makes sense for you?

@danielhuppmann  this looks fine for me, from my point of view can be merged

PR #51, opened 2020-06-15
Residential Demand Flexibility variables for 10 household technologies.

omnipotent12:
For Case Study 1. This is the continuation of #31 with all technologies considered and taking into account feedback and internal discussions within CS1.

### DIFF
--- a/nomenclature/definitions/variable/technology/technologies.yaml
+++ b/nomenclature/definitions/variable/technology/technologies.yaml
@@ -538,83 +538,83 @@ Investment|Energy Supply|Liquids|Oil:
    
    # description of residential demand flexibility potentials for 10 household technologies - potential curtailable load
    
-Maximum Curtailment|Electricity|Load Shifting|Residential|Refrigeration:
-   description: Maximum curtailable load due to changing household freezer/refrigerator usage.
+Demand Response|Maximum Reduction|Load Shifting|Electricity|Residential|Refrigeration:
+   description: Maximum load reduction due to changing household freezer/refrigerator usage.
    unit: MW
    
-Maximum Curtailment|Electricity|Load Shifting|Residential|Washing Machine:
-   description: Maximum curtailable load due to changing household washing machine usage.
+Demand Response|Maximum Reduction|Load Shifting|Electricity|Residential|Washing Machine:
+   description: Maximum load reduction due to changing household washing machine usage.
    unit: MW
    
-Maximum Curtailment|Electricity|Load Shifting|Residential|Dryer:
-   description: Maximum curtailable load due to changing household tumble dryer usage.
+Demand Response|Maximum Reduction|Load Shifting|Electricity|Residential|Dryer:
+   description: Maximum load reduction due to changing household tumble dryer usage.
    unit: MW
    
-Maximum Curtailment|Electricity|Load Shifting|Residential|Dish Washer:
-   description: Maximum curtailable load due to changing household dish washer usage.
+Demand Response|Maximum Reduction|Load Shifting|Electricity|Residential|Dish Washer:
+   description: Maximum load reduction due to changing household dish washer usage.
    unit: MW
 
-Maximum Curtailment|Electricity|Load Shifting|Residential|Air Conditioning:
-   description: Maximum curtailable load due to changing household air conditioning usage.
+Demand Response|Maximum Reduction|Load Shifting|Electricity|Residential|Air Conditioning:
+   description: Maximum load reduction due to changing household air conditioning usage.
    unit: MW
    
-Maximum Curtailment|Electricity|Load Shifting|Residential|Water Heater:
-   description: Maximum curtailable load due to changing household water storage heater usage.
+Demand Response|Maximum Reduction|Load Shifting|Electricity|Residential|Water Heater:
+   description: Maximum load reduction due to changing household water storage heater usage.
    unit: MW
    
-Maximum Curtailment|Electricity|Load Shifting|Residential|Space Heater:
-   description: Maximum curtailable load due to changing household space heater usage.
+Demand Response|Maximum Reduction|Load Shifting|Electricity|Residential|Space Heater:
+   description: Maximum load reduction due to changing household space heater usage.
    unit: MW
    
-Maximum Curtailment|Electricity|Load Shifting|Residential|Heat Pump:
-   description: Maximum curtailable load due to changing household heat circulation pump usage.
+Demand Response|Maximum Reduction|Load Shifting|Electricity|Residential|Heat Pump:
+   description: Maximum load reduction due to changing household heat circulation pump usage.
    unit: MW
    
-Maximum Curtailment|Electricity|Load Shifting|Residential|Electric Vehicle:
-   description: Maximum curtailable load due to changing household electric vehicle charging and usage.
+Demand Response|Maximum Reduction|Load Shifting|Electricity|Residential|Electric Vehicle:
+   description: Maximum load reduction due to changing household electric vehicle charging and usage.
    unit: MW
    
-Maximum Curtailment|Electricity|Load Shifting|Residential|Cooking:
-   description: Maximum curtailable load due to changing household electric stove, oven and microwave usage.
+Demand Response|Maximum Reduction|Load Shifting|Electricity|Residential|Cooking:
+   description: Maximum load reduction due to changing household electric stove, oven and microwave usage.
    unit: MW
    
    # description of residential demand flexibility potentials for 10 household technologies - limits of dispatchable load
-Maximum Dispatch|Electricity|Load Shifting|Residential|Refrigeration:
+Demand Response|Maximum Dispatch|Load Shifting|Electricity|Residential|Refrigeration:
    description: Maximum dispatchable load due to changing household freezer/refrigerator usage.
    unit: MW
 
-Maximum Dispatch|Electricity|Load Shifting|Residential|Washing Machine:
+Demand Response|Maximum Dispatch|Load Shifting|Electricity|Residential|Washing Machine:
    description: Maximum dispatchable load due to changing household washing machine usage.
    unit: MW
 
-Maximum Dispatch|Electricity|Load Shifting|Residential|Dryer:
+Demand Response|Maximum Dispatch|Load Shifting|Electricity|Residential|Dryer:
    description: Maximum dispatchable load due to changing household tumble dryer usage.
    unit: MW
    
-Maximum Dispatch|Electricity|Load Shifting|Residential|Dish Washer:
+Demand Response|Maximum Dispatch|Load Shifting|Electricity|Residential|Dish Washer:
    description: Maximum dispatchable load due to changing household dish washer usage.
    unit: MW
 
-Maximum Dispatch|Electricity|Load Shifting|Residential|Air Conditioning:
+Demand Response|Maximum Dispatch|Load Shifting|Electricity|Residential|Air Conditioning:
    description: Maximum dispatchable load due to changing household air conditioning usage.
    unit: MW
    
-Maximum Dispatch|Electricity|Load Shifting|Residential|Water Heater:
+Demand Response|Maximum Dispatch|Load Shifting|Electricity|Residential|Water Heater:
    description: Maximum dispatchable load due to changing household water storage heater usage.
    unit: MW
    
-Maximum Dispatch|Electricity|Load Shifting|Residential|Space Heater:
+Demand Response|Maximum Dispatch|Load Shifting|Electricity|Residential|Space Heater:
    description: Maximum dispatchable load due to changing household space heater usage.
    unit: MW
    
-Maximum Dispatch|Electricity|Load Shifting|Residential|Heat Pump:
+Demand Response|Maximum Dispatch|Load Shifting|Electricity|Residential|Heat Pump:
    description: Maximum dispatchable load due to changing household heat circulation pump usage.
    unit: MW
    
-Maximum Dispatch|Electricity|Load Shifting|Residential|Electric Vehicle:
+Demand Response|Maximum Dispatch|Load Shifting|Electricity|Residential|Electric Vehicle:
    description: Maximum dispatchable load due to changing household electric vehicle charging and usage.
    unit: MW
    
-Maximum Dispatch|Electricity|Load Shifting|Residential|Cooking:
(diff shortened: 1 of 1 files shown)


sandrinecharousset:
@omnipotent12  thank you for including those variables
Just to be sure our understanding is correct:
"potential' is an hourly profile corresponding to the hourly amount of load that can be shifted?
I am not completely sure to understand what Limit is. nadia sait it is the maximum consumption of the equipment (if eg the air conditionning is at maximum power)?
Do we not need a variable for stating the maximum energy (MWh) that can be shifted?
We also need the 'return to 0 duration' , that we understood shoould be depending on uses (number of hours in which the 'shifting can be done, with a 0-energy balance'); How could wa neme this?
I also have a comment regarding the order of 'words'
In order to be consistant with what is done for other technologies, I would propose to change a bit the order
Potential|Electricity|Load Shifting|Residential|<Use>
Limit|Electricity|Load Shifting|Residential|<Use>
Normal Return Duration|Electricity|Load Shifting|Residential|<Use>
(last one the name probably needs to be improved)

nomenclature/definitions/variable/technology/technologies.yaml
```diff
@@ -535,3 +535,106 @@ Investment|Energy Supply|Liquids|Oil:
                 capturing equipment should be included but not the one on
                 transport and storage.
    unit: billion US$2010/yr
+   
+   # description of residential demand flexibility potentials for 10 household technologies - potential curtailable load
+   
+Potential|Electricity|Load Shifting|Residential|Refrigeration:
+   description: Maximum curtailable load in a given region/time frame,
```
danielhuppmann, line 542: it is not necessary to specify "in a given region/time frame", this should be obvious anyway by having a region and a subannual column - and it removes the clutter from the descriptions

omnipotent12: ok, will change now, thanks!

danielhuppmann:
I think that "Limit" and "Potential" is not very obvious... If I understand it correctly, this is basically upwards and downwards changes to the load by shifting in time?

Maybe "Maximum Curtailment" - and whatever is the opposite of Curtailment...

omnipotent12:
> I think that "Limit" and "Potential" is not very obvious... If I understand it correctly, this is basically upwards and downwards changes to the load by shifting in time?
> 
> Maybe "Maximum Curtailment" - and whatever is the opposite of Curtailment...
@danielhuppmann, thanks for the idea! The changes have been implemented in the latest Committ.

danielhuppmann:
"Maximum Dispatch" is clearer - @sandrinecharousset what do you think?

sandrinecharousset:
@danielhuppmann  @omnipotent12 I would not use the word curtailement regarding load shifting because we also model another kind of demand response which is called load curtailement. (load shifting consists in moving load from one time slice to another, with 0 balance regarding energy, while load curtailement means erasing some of the load, with a negative balance in energy, and constraints regarding the total amount of energy that can be curtailed

omnipotent12:
> 
> 
> @omnipotent12 thank you for including those variables
> Just to be sure our understanding is correct:
> "potential' is an hourly profile corresponding to the hourly amount of load that can be shifted?
> I am not completely sure to understand what Limit is. nadia sait it is the maximum consumption of the equipment (if eg the air conditionning is at maximum power)?
> Do we not need a variable for stating the maximum energy (MWh) that can be shifted?
> We also need the 'return to 0 duration' , that we understood shoould be depending on uses (number of hours in which the 'shifting can be done, with a 0-energy balance'); How could wa neme this?
> I also have a comment regarding the order of 'words'
> In order to be consistant with what is done for other technologies, I would propose to change a bit the order
> Potential|Electricity|Load Shifting|Residential|
> Limit|Electricity|Load Shifting|Residential|
> Normal Return Duration|Electricity|Load Shifting|Residential|
> (last one the name probably needs to be improved)

Thank you for the discussion @sandrinecharousset! Our understanding of the variables is the same, and I have implemented your suggested changes. 

I would suggest we leave the maximum energy and return duration variables out of the nomenclature for now to reduce clutter and add them later if needed. Some potential variables (e.g. return duration) I see more as parameters assumed within the models that may not need to be exchanged and viewed on the Platform; they can instead perhaps define scenarios in CS1.

nomenclature/definitions/variable/technology/technologies.yaml
```diff
@@ -538,103 +538,83 @@ Investment|Energy Supply|Liquids|Oil:
    
    # description of residential demand flexibility potentials for 10 household technologies - potential curtailable load
    
-Load Shifting Potential|Electricity|Residential|Refrigeration:
-   description: Maximum curtailable load in a given region/time frame,
-                due to changing household freezer/refrigerator usage.
+Maximum Curtailment|Electricity|Load Shifting|Residential|Refrigeration:
```
sandrinecharousset, line 541: Curtailement should not be used as it means something different (see global comment)

nomenclature/definitions/variable/technology/technologies.yaml
```diff
@@ -538,103 +538,83 @@ Investment|Energy Supply|Liquids|Oil:
...
-                due to changing household electric stove, oven and microwave usage.
+Maximum Curtailment|Electricity|Load Shifting|Residential|Cooking:
+   description: Maximum curtailable load due to changing household electric stove, oven and microwave usage.
    unit: MW
    
    # description of residential demand flexibility potentials for 10 household technologies - limits of dispatchable load
-   Load Shifting Limit|Electricity|Residential|Refrigeration:
-   description: Maximum dispatchable load in a given region/time frame,
-                due to changing household freezer/refrigerator usage.
+Maximum Dispatch|Electricity|Load Shifting|Residential|Refrigeration:
```
sandrinecharousset, line 582: Maximum Dispatch is fine

nomenclature/definitions/variable/technology/technologies.yaml
omnipotent12: "Maximum Reduction"?

sandrinecharousset:
> > @omnipotent12 thank you for including those variables
> > Just to be sure our understanding is correct:
> > "potential' is an hourly profile corresponding to the hourly amount of load that can be shifted?
> > I am not completely sure to understand what Limit is. nadia sait it is the maximum consumption of the equipment (if eg the air conditionning is at maximum power)?
> > Do we not need a variable for stating the maximum energy (MWh) that can be shifted?
> > We also need the 'return to 0 duration' , that we understood shoould be depending on uses (number of hours in which the 'shifting can be done, with a 0-energy balance'); How could wa neme this?
> > I also have a comment regarding the order of 'words'
> > In order to be consistant with what is done for other technologies, I would propose to change a bit the order
> > Potential|Electricity|Load Shifting|Residential|
> > Limit|Electricity|Load Shifting|Residential|
> > Normal Return Duration|Electricity|Load Shifting|Residential|
> > (last one the name probably needs to be improved)
> 
> Thank you for the discussion @sandrinecharousset! Our understanding of the variables is the same, and I have implemented your suggested changes.
> 
> I would suggest we leave the maximum energy and return duration variables out of the nomenclature for now to reduce clutter and add them later if needed. Some potential variables (e.g. return duration) I see more as parameters assumed within the models that may not need to be exchanged and viewed on the Platform; they can instead perhaps define scenarios in CS1.

@omnipotent12 this makes sense (leaving return time out); It can be considered as model parameters

sandrinecharousset:
> I think that "Limit" and "Potential" is not very obvious... If I understand it correctly, this is basically upwards and downwards changes to the load by shifting in time?
> 
> Maybe "Maximum Curtailment" - and whatever is the opposite of Curtailment...

@omnipotent12 @danielhuppmann 
To my point of view, limit was not obvious but I think potential is not so bad (we always talk of load shifting potential....); maybe capacity could do it?

danielhuppmann:
Good point @sandrinecharousset about the different meaning of curtailment.

How about a different structure:

`Demand Response|Electricity|Curtailment|Potential|...`
`Demand Response|Electricity|Load Shifting|Potential|Increase|...`
`Demand Response|Electricity|Load Shifting|Potential|Reduction|...`

`Demand Response|Electricity|Curtailment|Actual|...`
`Demand Response|Electricity|Load Shifting|Actual|Increase|...`
`Demand Response|Electricity|Load Shifting|Actual|Reduction|...`

nomenclature/definitions/variable/technology/technologies.yaml
sandrinecharousset: Maximum reduction could do it but maybe we would need maximum reduction of sth?
Why not capacity? (load shifting is already in the variable name)?

sandrinecharousset:
> Good point @sandrinecharousset about the different meaning of curtailment.
> 
> How about a different structure:
> 
> `Demand Response|Electricity|Curtailment|Potential|...`
> `Demand Response|Electricity|Load Shifting|Potential|Increase|...`
> `Demand Response|Electricity|Load Shifting|Potential|Reduction|...`
> 
> `Demand Response|Electricity|Curtailment|Actual|...`
> `Demand Response|Electricity|Load Shifting|Actual|Increase|...`
> `Demand Response|Electricity|Load Shifting|Actual|Reduction|...`

@danielhuppmann , Actual and Potential do not really make sense. In particular Actual load shifting does not exist. The idea is that there is an equipment that is consuming energy, it has one 'forecasted' consumption profile. This  profile can be changed (and these are the decisions taken by odels): either increase its capacity, thus increasing the load (this is the maximum dispatch, former limit), or you can decrease it in some limit (thus it is equivalent to virtual generation), which is what is meant by 'potential'. Within load shifting, the energy balance on a certain time period has to be zero (this is why we may need pther variables related to that 'time window', which we are thinking of leaving out of the nomenclature as they are more 'parameters'. The distinction 'actual' 'potential' thus is not needed. (the actual mode is not doing anything)

danielhuppmann:
But how would you describe the difference between the planned load profile and the realized load? There should be a variable for that - if only to calculate how much a consumer is paid for the load-shifting...

sandrinecharousset:
> > Good point @sandrinecharousset about the different meaning of curtailment.
> > How about a different structure:
> > `Demand Response|Electricity|Curtailment|Potential|...`
> > `Demand Response|Electricity|Load Shifting|Potential|Increase|...`
> > `Demand Response|Electricity|Load Shifting|Potential|Reduction|...`
> > `Demand Response|Electricity|Curtailment|Actual|...`
> > `Demand Response|Electricity|Load Shifting|Actual|Increase|...`
> > `Demand Response|Electricity|Load Shifting|Actual|Reduction|...`
> 
> @danielhuppmann , Actual and Potential do not really make sense. In particular Actual load shifting does not exist. The idea is that there is an equipment that is consuming energy, it has one 'forecasted' consumption profile. This profile can be changed (and these are the decisions taken by odels): either increase its capacity, thus increasing the load (this is the maximum dispatch, former limit), or you can decrease it in some limit (thus it is equivalent to virtual generation), which is what is meant by 'potential'. Within load shifting, the energy balance on a certain time period has to be zero (this is why we may need pther variables related to that 'time window', which we are thinking of leaving out of the nomenclature as they are more 'parameters'. The distinction 'actual' 'potential' thus is not needed. (the actual mode is not doing anything)

@danielhuppmann why having demand response at the beginning? For me demand response is a category of 'technologies' suc as 'storage' 'fossils' 'res'; load shifting and load curtailment are specific demand-reponse 'technologies', then maximum dispatch and limit (the new name we are going to find for limit) are the variables defining the technology (such as capacity, maximum volume.... for generation/storage technologies)
In the results there will be some output variables defining the 'dispatch' of these 'units' that could be names 'Active power' (as shifting a load is from the point of view of the electricity system equivalent to generating energy, in some cases with negative emissions)

sandrinecharousset:
> But how would you describe the difference between the planned load profile and the realized load? There should be a variable for that - if only to calculate how much a consumer is paid for the load-shifting...

This is in the 'demand' side of the variables. Sth. like Final Energy|....|Refrigeration
But we do not need to know the planned load for each specific use. We ony need to know what can be erased

omnipotent12:
Thanks for the discussion! I suggest we just add the variables we know we will need on the platform for the moment and revisit this if we find out others are necessary. Many potential variables may stay on the model side and not need to be exhanged/viewed on the platform. 

I would put forth the following:
`Maximum Reduction|Electricity|Load Shifting|Residential|<technology>`
and
`Maximum Dispatch|Electricity|Load Shifting|Residential|<technology>`

The term `Load Shifting` can be juxtaposed with `Curtailment` if we end up differentiating, and we can add a `Curtailment` suite of variables then. 

Sound good @sandrinecharousset, @danielhuppmann ?

sandrinecharousset:
> t

this sounds good for me
I will add variables related to load curtailment as they are neede by the plan4eu model (but not -yet?- differentiated per technologies), as we need to include it (in some countries this is active, eg in france with a quite big volume of curtailment, and we can fill those variables from our expertise)

danielhuppmann:
You don't like putting it under a broad "Demand Response" category?

sandrinecharousset:
> You don't like putting it under a broad "Demand Response" category?

yes would make sense
Maximum Reduction|Electricity|Demand response|Load Shifting|Residential|<technology>
and
Maximum Dispatch|Electricity|Demand response|Load Shifting|Residential|<technology>
@omnipotent12 ?

omnipotent12:
> 
> 
> > You don't like putting it under a broad "Demand Response" category?
> 
> yes would make sense
> Maximum Reduction|Electricity|Demand response|Load Shifting|Residential|
> and
> Maximum Dispatch|Electricity|Demand response|Load Shifting|Residential|
> @omnipotent12 ?

This works for me too! @danielhuppmann does this preserve the spirit of what you proposed?

danielhuppmann:
> @danielhuppmann does this preserve the spirit of what you proposed?

I was thinking about having Demand Response as the top-level category, because dispatch and reduction are specifications of how demand responds...

omnipotent12:
> 
> I was thinking about having Demand Response as the top-level category, because dispatch and reduction are specifications of how demand responds...

Ok @danielhuppmann, I would propose the following then:
`Demand Response|Maximum Reduction|Load Shifting|Electricity|Residential|<technology>`
and
`Demand Response|Maximum Dispatch|Load Shifting|Electricity|Residential|<technology>`
The first three terms specify the type of demand response, and the last three terms specify the the sector.

omnipotent12:
@danielhuppmann all changes should be implemented, please let me know if anything else is needed. Thanks!

PR #56, opened 2020-06-24
Addition of the continuous-time format

erikfilias:
Main details about the continuous-time format (daily and hourly data):

- [x] Standard datetime format     : **UTC**  
> For example: `2020-01-01, 2020-01-02, ...` or `2020-01-01T13:00, ...`
> Inclusion of attributes is possible such as duration (More details [here](https://github.com/openENTRANCE/nomenclature/blob/master/nomenclature/definitions/subannual/months.yaml))
- [x] Using time zone                    : **no relevance**

Then, this PR considers an update of the README related to subannual section, and scripts to get a list of dates and hours, respectively.
##
The format and details for representative timeslices will be performed in a new PR, taking in account the following:

- [x] Using winter/summer time   : 
> To distinguish between different granularity levels of representative timeslices, It was proposed the following: `<Granularity>|<Name of timeslice>`. For example: `2 Season-2 Times|Summer-Day`.
- [x] Averaging values over the time span   : 
> A value is always the average for flow variables (i.e. It is the average between the subannual time and the subsequent one, where average is contingent on the lowest level of granularity - if you use 2020-01-01T13:00, it is hourly average, if you use 2020-01-01, it is daily average...). 
> Reference comment -> [here](https://github.com/openENTRANCE/nomenclature/issues/46#issuecomment-641721712)
- [x] Accumulating values over the time span: 
> A value at the start of the period for stock over the time-period until the start of the next timeslice. 
> For example, Capacity or Reservoir Level at 2020-01-01T13:00 is the value at 1pm, if subannual is given as 2020-01-01, it is understood as midnight that day, if it's January, it is understood as midnight on the first day of the month. 
> Reference comment -> [here](https://github.com/openENTRANCE/nomenclature/issues/46#issuecomment-641721712)


nomenclature/definitions/subannual/README.md
```diff
@@ -43,6 +43,60 @@ with open(f'../subannual/months.yaml', 'r') as stream:
...
 
+### Daily and hourly data
+Our convention makes reference to the following agreements:
+
+
+#### UTC datetime format:
+- Standard datetime format     : **UTC**  
+> The UTC datetime format is adopted to consider the following levels: year, month, day, hour, minute, and second (without consider time zones).
+> For example: `2020-01-01, 2020-01-02, ...` or `2020-01-01T13:00, ...`
+> Inclusion of attributes is possible such as duration (More details [here](https://github.com/openENTRANCE/nomenclature/blob/master/nomenclature/definitions/subannual/months.yaml))
```
sandrinecharousset, line 54: The more details link opens a file describing months?

nomenclature/definitions/subannual/README.md
```diff
@@ -43,6 +43,59 @@ with open(f'../subannual/months.yaml', 'r') as stream:
...
+from datetime import timedelta, datetime
+
+def DateTimeRange(datetime1, datetime2):
+    for n in range(int (((datetime2 - datetime1).days)+1)*24):
+        yield datetime1 + timedelta(hours=n)
+
+start_dt = datetime(2020, 6, 23)
+end_dt   = datetime(2020, 6, 30)
+for dt in DateTimeRange(start_dt, end_dt):
+    print(dt.strftime("%Y-%m-%d %H:%M:%S"))
```
danielhuppmann, line 86: This snippet is not consistent with the example given above!

nomenclature/definitions/subannual/README.md
```diff
@@ -43,6 +43,59 @@ with open(f'../subannual/months.yaml', 'r') as stream:
 mapping = dict([(m, eval(attr['duration'])) for (m, attr) in months.items()])
 ```
 
+### Daily and hourly data
+Our convention makes reference to the following agreements:
+
```
danielhuppmann, line 48: Please only use one empty line in between headers and paragraphs
```suggestion
```

nomenclature/definitions/subannual/README.md
```diff
@@ -43,6 +43,59 @@ with open(f'../subannual/months.yaml', 'r') as stream:
...
+
+#### UTC datetime format:
+- Standard datetime format     : **UTC**  
+> The UTC datetime format is adopted to consider the following levels: year, month, day, hour, minute, and second (without consider time zones).
+> For example: `2020-01-01, 2020-01-02, ...` or `2020-01-01T13:00, ...`
+> The use of lists of dates and datetimes will be generated by the codes of the next section.
+- Using time zone                    : **no relevance**
```
danielhuppmann, line 55: I think writing "no relevance" is misleading - this might be clearly relevant for some contexts, but we should rather explain that we do not include special treatment for identifying time zones and assume everything is in UTC-Z.

nomenclature/definitions/subannual/README.md
```diff
@@ -43,6 +43,59 @@ with open(f'../subannual/months.yaml', 'r') as stream:
...
+    for n in range(int (((datetime2 - datetime1).days)+1)*24):
+        yield datetime1 + timedelta(hours=n)
+
+start_dt = datetime(2020, 6, 23)
+end_dt   = datetime(2020, 6, 30)
+for dt in DateTimeRange(start_dt, end_dt):
+    print(dt.strftime("%Y-%m-%d %H:%M:%S"))
+```
+
+The format `"%Y-%m-%d %H:%M:%S"` is composed by tokens. Each token represents a different part of the date-time, like day, month, year, etc. More details can be found in [strftime() and strptime() section](https://docs.python.org/3/library/datetime.html).
```
danielhuppmann, line 89: There is no such section yet, right?

erikfilias: Yeah, right. It should be written "strftime() and strptime() Format Codes".

nomenclature/definitions/subannual/README.md
erikfilias: OK, I got it.

sebastianzwickl:
From our (@tperger) point of view, the current representation of the months is not intuitive and consistent with the extension in this PR regarding the hourly and weekly resolution of the time series. The monthly descriptions (January, February, ...) are currently entered in the column "Subannual". We suggest instead **Month in this column**. The columns with the UTC time format will then not differ between day, week and month. Examples to make it clearer:

- Month

 | **...** | **region** | **variable**   | **unit** | **subannual** | **2020-01-01** | **2020-02-01** | **2020-03-01** |
|--------------|------------|----------------|----------|---------------|---------:|---------:|---------:|
 | ... | World      | Primary Energy | EJ/y     | Month|    462.5 |    500.7 |      522.4 |

- Week

 | **...** | **region** | **variable**   | **unit** | **subannual** | **2020-01-01** | **2020-01-08** | **2020-01-15** |
|--------------|------------|----------------|----------|---------------|---------:|---------:|---------:|
 | ... | World      | Primary Energy | EJ/y     | Weekly|    462.5 |    500.7 |      522.4 |

- Day

 | **...** | **region** | **variable**   | **unit** | **subannual** | **2020-01-01** | **2020-01-02** | **2020-01-3** |
|--------------|------------|----------------|----------|---------------|---------:|---------:|---------:|
 | ... | World      | Primary Energy | EJ/y     | Daily|    462.5 |    500.7 |      522.4 |

- Hour

 | **...** | **region** | **variable**   | **unit** | **subannual** | **2020-01-01 00:00:00** | **2020-01-01 01:00:00** | **2020-01-15 02:00:00** |
|--------------|------------|----------------|----------|---------------|---------:|---------:|---------:|
 | ... | World      | Primary Energy | EJ/y     | Hourly|    462.5 |    500.7 |      522.4 |

erikfilias:
> From our (@tperger) point of view, the current representation of the months is not intuitive and consistent with the extension in this PR regarding the hourly and weekly resolution of the time series. The monthly descriptions (January, February, ...) are currently entered in the column "Subannual". We suggest instead **Monthly in this column**. The columns with the UTC time format will then not differ between day, week and month. Examples to make it clearer:
> 
> * Month
> 
> **...**	**region**	**variable**	**unit**	**subannual**	**2020-01-01**	**2020-02-01**	**2020-03-01**
> ...	World	Primary Energy	EJ/y	Monthly	462.5	500.7	522.4
> * Week
> 
> **...**	**region**	**variable**	**unit**	**subannual**	**2020-01-01**	**2020-01-08**	**2020-01-15**
> ...	World	Primary Energy	EJ/y	Weekly	462.5	500.7	522.4
> * Day
> 
> **...**	**region**	**variable**	**unit**	**subannual**	**2020-01-01**	**2020-01-02**	**2020-01-3**
> ...	World	Primary Energy	EJ/y	Daily	462.5	500.7	522.4
> * Hour
> 
> **...**	**region**	**variable**	**unit**	**subannual**	**2020-01-01 00:00:00**	**2020-01-01 01:00:00**	**2020-01-15 02:00:00**
> ...	World	Primary Energy	EJ/y	Hourly	462.5	500.7	522.4

I agree in part with you about **Monthly in column**. In addition, we could maintain the UTC time format in "Subannual" in the representation of the months. Cause in a daily or hourly representation we could have several columns.

tperger:
You are right @erikfilias, we might end up with many columns. But imo having time stamps in the rows _and_ columns is rather confusing. It would be more intuitive to have one line per time series and the subannual columns telling us what it is (e.g. the resolution)

danielhuppmann:
In my opinion, there are three concerns:

 - what is intuitive to the user: seeing a column value "January", it is immediately obvious to a human that this is monthly data - and this also holds for other disaggregation levels
 - providing functionality to support automated (dis)aggregation, making the hierarchy of temporal resolution easily digestible to a workflow (e.g., how does a script know that `01-01-2020` is a member of `January`)
 - how to inform a (human) user and a script what the resolution of a model is (e.g., every 4 hours - would the "Subannual" column state "Hourly"?)

I don't see how this proposal increases intuition or improves the third point...

erikfilias:
You right @danielhuppmann . We could include a TimeStep parameter in our data as we made in openTEPES. Cause at first, we start with hourly resolution for a year, and after that a the LoadLevels (timestamps) are averaged automatically according to the TimeStep (e.g., every 2 hours).

@tperger , you right  too. This idea of using columns could be used for representative timeslices to reduce the rows in the file, I think.

sebastianzwickl:
> In my opinion, there are three concerns:
> 
> * what is intuitive to the user: seeing a column value "January", it is immediately obvious to a human that this is monthly data - and this also holds for other disaggregation levels
>
Just so I understand: What is the intuitive equivalent of "January" for the first day or week of the year? (This might be a question at the other levels of aggregation mentioned above I guess.)
> * how to inform a (human) user and a script what the resolution of a model is (e.g., every 4 hours - would the "Subannual" column state "Hourly"?)
>
As far as I understood it the attribute `duration` is intended for this. One could also use the pipe structure (as with the variables) and use `Hourly|4h.`

sandrinecharousset:
> > In my opinion, there are three concerns:
> > 
> > * what is intuitive to the user: seeing a column value "January", it is immediately obvious to a human that this is monthly data - and this also holds for other disaggregation levels
> 
> Just so I understand: What is the intuitive equivalent of "January" for the first day or week of the year? (This might be a question at the other levels of aggregation mentioned above I guess.)
> 
> > * how to inform a (human) user and a script what the resolution of a model is (e.g., every 4 hours - would the "Subannual" column state "Hourly"?)
> 
> As far as I understood it the attribute `duration` is intended for this. One could also use the pipe structure (as with the variables) and use `Hourly|4h.`

Regarding the weeks, should not we include week1, week2 (or W1, W2) like what is used in our favourite calendars (I do not know your uses, but here we often relate to W23 as the 23rd week of the year, and the definition is the same for everyone; weel1 is the week which comprises january1st . 
Regarding days, I believe UTC is enough 2020-06-30 is understood as a day by everyone...

sandrinecharousset:
> From our (@tperger) point of view, the current representation of the months is not intuitive and consistent with the extension in this PR regarding the hourly and weekly resolution of the time series. The monthly descriptions (January, February, ...) are currently entered in the column "Subannual". We suggest instead **Monthly in this column**. The columns with the UTC time format will then not differ between day, week and month. Examples to make it clearer:
> 
> * Month
> 
> **...**	**region**	**variable**	**unit**	**subannual**	**2020-01-01**	**2020-02-01**	**2020-03-01**
> ...	World	Primary Energy	EJ/y	Monthly	462.5	500.7	522.4
> * Week
> 
> **...**	**region**	**variable**	**unit**	**subannual**	**2020-01-01**	**2020-01-08**	**2020-01-15**
> ...	World	Primary Energy	EJ/y	Weekly	462.5	500.7	522.4
> * Day
> 
> **...**	**region**	**variable**	**unit**	**subannual**	**2020-01-01**	**2020-01-02**	**2020-01-3**
> ...	World	Primary Energy	EJ/y	Daily	462.5	500.7	522.4
> * Hour
> 
> **...**	**region**	**variable**	**unit**	**subannual**	**2020-01-01 00:00:00**	**2020-01-01 01:00:00**	**2020-01-15 02:00:00**
> ...	World	Primary Energy	EJ/y	Hourly	462.5	500.7	522.4

I thought we had decided not to add columns? (only the subannual one) and treat all data in different lines to keep consistancy. But I may have missed sth....

sandrinecharousset:
> @danielhuppmann . We could include a TimeStep parameter in our data as we made in openTEPES. Cause at first, we start with hourly resolution for a year, and after that a the LoadLevels (timestamps) are averaged automatically according to the TimeStep (e.g., every 2 hours).
> 
> @tperger , you right. This idea of using columns could be used for representative timeslices to reduce the rows in the file, I think.

I should think that kind of parameter is not a data but a model parameter and probably should not be included in the data, no?

danielhuppmann:
> I thought we had decided not to add columns? (only the subannual one) and treat all data in different lines to keep consistancy. But I may have missed sth....

Indeed, there was no decision to change the columns, and I also do not see any convincing argument why a change would be better (more intuitive and/or easier to process).

sandrinecharousset:
> > From our (@tperger) point of view, the current representation of the months is not intuitive and consistent with the extension in this PR regarding the hourly and weekly resolution of the time series. The monthly descriptions (January, February, ...) are currently entered in the column "Subannual". We suggest instead **Monthly in this column**. The columns with the UTC time format will then not differ between day, week and month. Examples to make it clearer:
> > 
> > * Month
> > 
> > **...**	**region**	**variable**	**unit**	**subannual**	**2020-01-01**	**2020-02-01**	**2020-03-01**
> > ...	World	Primary Energy	EJ/y	Monthly	462.5	500.7	522.4
> > 
> > * Week
> > 
> > **...**	**region**	**variable**	**unit**	**subannual**	**2020-01-01**	**2020-01-08**	**2020-01-15**
> > ...	World	Primary Energy	EJ/y	Weekly	462.5	500.7	522.4
> > 
> > * Day
> > 
> > **...**	**region**	**variable**	**unit**	**subannual**	**2020-01-01**	**2020-01-02**	**2020-01-3**
> > ...	World	Primary Energy	EJ/y	Daily	462.5	500.7	522.4
> > 
> > * Hour
> > 
> > **...**	**region**	**variable**	**unit**	**subannual**	**2020-01-01 00:00:00**	**2020-01-01 01:00:00**	**2020-01-15 02:00:00**
> > ...	World	Primary Energy	EJ/y	Hourly	462.5	500.7	522.4
> 
> I agree in part with you about **Monthly in column**. In addition, we could maintain the UTC time format in "Subannual" in the representation of the months. Cause in a daily or hourly representation we could have several columns. For example, I'm planning to include data about 8760 hours.

I am also going to incude hourly data (8760 per year then), but I had planned to have one line per hour, so that we can easily use the aggregation/disaggregation functions.... and morover I doubt we can host files with 8766 columns.... (where files with too many lines can easily be cut into pieces)

sandrinecharousset:
> > I thought we had decided not to add columns? (only the subannual one) and treat all data in different lines to keep consistancy. But I may have missed sth....
> 
> Indeed, there was no decision to change the columns, and I also do not see any convincing argument why a change would be better (more intuitive and/or easier to process).

I personnaly see arguments against... (all my data are organised with 1 hour per line, changing this would necessitate more transformations - not complex but as the volumetry of the data is huge, long to process.... plus the fact that file with 8766 columns are not handable.....). (thanks for your answer, I was a bit worried having to do all this....)

danielhuppmann:
Responding to @sebastianzwickl 

> As far as I understood it the attribute duration is intended for this.

The duration attribute can be included where it is obvious and where it can be helpful in the automated use of (dis)aggregation tools in the post-processing (e.g., take all timeslices that have the attribute `category='Month'` and do `sum of value of month times duration of month`). If choosing option B (below), giving a `duration` attribute for UTC timestamps is not feasible.

> One could also use the pipe structure (as with the variables) and use Hourly|4h.

This is the crux of the matter: say you have `Model A` that has hourly resolution and `Model B` that has `4-hourly` resolution.

**Option 1:** You can define two sets of subannual timesteps, where `Model A` submits all data as `Hourly|2020-01-01T01:00` and `Model B` submits data as `4-Hourly|2020-01-01T01:00` - and you provide a mapping between the two.

**Option 2:** Both models provide their data directly in UTC-compatible format `2020-01-01T01:00` and the workflow for comparing the results or using one as input to the other has to be structured such that it knows/discovers that `Model B` only provides every fourth data point.

danielhuppmann:
And to make matters more complicated (I'm sorry): @erikfilias and @sandrinecharousset are going to submit data at a resolution of 8760 hours per year. Is `Hour 0 = January 1 midnight until 1am` assumed to be a Monday? Because (UTC-style) `2020-01-01T00:00` was actually a Wednesday... And `2021-01-01T00:00` will be a Friday... This obviously has repercussion on the load profiles.

So we should find an agreement whether we follow the actual calendar years in using UTC (with leapers?), or if we construct a simplified "hypothetical" archetype year.

arght:
Regarding last comment from Daniel there is no problem for us to use UTC style to enumerate the 8760 hours of a year and internally the model is not affected by a year starting on Wednesday, Friday or whatsoever. In any case it should be responisbility of the modeler to take this into account and to make it consistent with the load, for example.

erikfilias:
> And to make matters more complicated (I'm sorry): @erikfilias and @sandrinecharousset are going to submit data at a resolution of 8760 hours per year. Is `Hour 0 = January 1 midnight until 1am` assumed to be a Monday? Because (UTC-style) `2020-01-01T00:00` was actually a Wednesday... And `2021-01-01T00:00` will be a Friday... This obviously has repercussion on the load profiles.
> 
> So we should find an agreement whether we follow the actual calendar years in using UTC (with leapyears?), or if we construct a simplified "hypothetical" archetype year.

In our case to assume a Monday, Tuesday,... for the `Hour 0` is indifferent. E.g., we consider the range from `2030-01-01T00:00` to `2030-12-30T23:00` and only matter if we consider a TimeStep. 
However, about using snapshots or representative timeslices, it could be crucial.
Please, correct me if I'm wrong, @arght.

sandrinecharousset:
> And to make matters more complicated (I'm sorry): @erikfilias and @sandrinecharousset are going to submit data at a resolution of 8760 hours per year. Is `Hour 0 = January 1 midnight until 1am` assumed to be a Monday? Because (UTC-style) `2020-01-01T00:00` was actually a Wednesday... And `2021-01-01T00:00` will be a Friday... This obviously has repercussion on the load profiles.
> 
> So we should find an agreement whether we follow the actual calendar years in using UTC (with leapyears?), or if we construct a simplified "hypothetical" archetype year.

The data I am going to include have been treated so to fit to the calendar of 2050; in particular regarding electricity loads, there are dependencies with the calendar: seasons, but also the profile of a week, specific profiles for week-days, week-end days, fridays are different also, and also bank holidays, and all school holidays of all countries.... 1/1/2050 is a saturday.
If I were to create a dataset for 2030 we would have to rerun all models again (and it is not only data treatment, but complete runs of electicity load models for all countries.....); knowing that we also have 50 scenarios (of 8760 values each), which represent the 'climatic -past but readapted to the climatic ambiance of 2050- years...
Then I definetely would not enjoy a forced 1/1 is a monday :-)

sandrinecharousset:
> > And to make matters more complicated (I'm sorry): @erikfilias and @sandrinecharousset are going to submit data at a resolution of 8760 hours per year. Is `Hour 0 = January 1 midnight until 1am` assumed to be a Monday? Because (UTC-style) `2020-01-01T00:00` was actually a Wednesday... And `2021-01-01T00:00` will be a Friday... This obviously has repercussion on the load profiles.
> > So we should find an agreement whether we follow the actual calendar years in using UTC (with leapyears?), or if we construct a simplified "hypothetical" archetype year.
> 
> The data I am going to include have been treated so to fit to the calendar of 2050; in particular regarding electricity loads, there are dependencies with the calendar: seasons, but also the profile of a week, specific profiles for week-days, week-end days, fridays are different also, and also bank holidays, and all school holidays of all countries.... 1/1/2050 is a saturday.
> If I were to create a dataset for 2030 we would have to rerun all models again (and it is not only data treatment, but complete runs of electicity load models for all countries.....); knowing that we also have 50 scenarios (of 8760 values each), which represent the 'climatic -past but readapted to the climatic ambiance of 2050- years...
> Then I definetely would not enjoy a forced 1/1 is a monday :-)

just to complement, we do not need to tell the model that 1/1/2050 is a saturday, there are plenty of nice functions that compute that easily.....
Same thing to know when Week1 is starting: the "rule" is that week1 is the week where jan1 is, which means Week 1 of 2020 starts december27, 2019 (so for 2020, it is a 2 days week)

danielhuppmann:
Ok, so there is a consensus that we use "real" calendar days. So for being specific, this implies that the years 2020, 2024, etc. have 8784 hours (not 8760).

@erikfilias can you please update the proposed changes to the README:
 - [ ] explain that "real" calendar days must be used and that the number of hours to be used in any hourly model resolution depends on whether a year is a leap year or not
 - [ ] explain how modellers should deal with the switch to summer-time (as a part of the agreement to use UTC)
 - [ ] clarify that all time data must be given in UTC (currently, it says "time zone: no relevance" - this is not helpful as I already wrote as an inline comment)
 - [ ] explain that the hour `2020-1-1T00:00` is understood as the duration from midnight until 1am (or whatever the next data point is in a multi-hour timestep resolution) and that it is the responsibility of the modelling team (this applies for example to openTEPES) must transform the data as part of the input/output processing workflow

danielhuppmann:
responding to @sandrinecharousset 

> Same thing to know when Week1 is starting: the "rule" is that week1 is the week where jan1 is, which means Week 1 of 2020 starts december27, 2019 (so for 2020, it is a 2 days week)

Can we please hold off on that discussion for a new PR after this one is merged? The discussion on how to deal with UTC and (multi-)hourly data seems already complicated enough...

sandrinecharousset:
> responding to @sandrinecharousset
> 
> > Same thing to know when Week1 is starting: the "rule" is that week1 is the week where jan1 is, which means Week 1 of 2020 starts december27, 2019 (so for 2020, it is a 2 days week)
> 
> Can we please hold off on that discussion for a new PR after this one is merged? The discussion on how to deal with UTC and (multi-)hourly data seems already complicated enough...

yes of course, anyway we already can deal with weekly data provided that we use UTC for dates and a duration of 168 hours....Not even sure any model would require to define weeks

danielhuppmann:
thanks @erikfilias, I'm confused... we previously always discussed to use UTC-format, but now, the description in the README says that we use Central European Time... So if a team uploads a timestamp `2020-01-01T00:00` - is that understood as midnight in London (UTC) or midnight in Berlin/Vienna/Paris...?

erikfilias:
> thanks @erikfilias, I'm confused... we previously always discussed to use UTC-format, but now, the description in the README says that we use Central European Time... So if a team uploads a timestamp `2020-01-01T00:00` - is that understood as midnight in London (UTC) or midnight in Berlin/Vienna/Paris...?

Yeah sorry, it should be **Everything is in the same time zone (Berlin, Vienna, Paris, Madrid, etc.)**. I corrected it.

Should I rebase the PR to include "stickler-ci" file?

danielhuppmann:
I think the worst thing to do when establishing a data-sharing convention is to come up with something that is *almost identical* to a very well-established format - in this case, you propose to use `2020-1-1T00:00` (which looks like UTC convention) to mean `2020-1-1T00:00+01:00`. This will inevitably lead to confusion, because users don't read the fine print in the docs.

Two options to resolve this:
 - we use proper UTC format (without specifying a time zone and the time has to be converted to GMT)
 - we specifically add the `+01:00` tag to the proposed format

erikfilias:
> I think the worst thing to do when establishing a data-sharing convention is to come up with something that is _almost identical_ to a very well-established format - in this case, you propose to use `2020-1-1T00:00` (which looks like UTC convention) to mean `2020-1-1T00:00+01:00`. This will inevitably lead to confusion, because users don't read the fine print in the docs.
> 
> Two options to resolve this:
> 
> * we use proper UTC format (without specifying a time zone and the time has to be converted to GMT)
> * we specifically add the `+01:00` tag to the proposed format

For me It's ok option 2, since we'll use the complete format. I'll try to include a clarified information about it in README if you agree.

danielhuppmann:
@openENTRANCE/coordination, do you agree that we should require to add the CET-timestamp `+01:00` (option 2 above)?

tburandt:
> @openENTRANCE/coordination, do you agree that we should require to add the CET-timestamp `+01:00` (option 2 above)?

Yes, I also agree on this.

nomenclature/definitions/subannual/README.md
```diff
@@ -43,6 +43,85 @@ with open(f'../subannual/months.yaml', 'r') as stream:
...
+
+- Using winter/summer time as representatives timeslices  :
+> To distinguish between different granularity levels of representative timeslices, It was proposed the following: `<Granularity>|<Name of timeslice>`. For example: `2 Season-2 Times|Summer-Day`.
+
+- Averaging values over the time span   :
+> A value is always the average for flow variables (i.e. It is the average between the subannual time and the subsequent one, where average is contingent on the lowest level of granularity - if you use 2020-01-01T13:00, it is hourly average, if you use 2020-01-01, it is daily average...).
+
+- Accumulating values over the time span:
+> A value at the start of the period for stock over the time-period until the start of the next timeslice.
+> For example, Capacity or Reservoir Level at 2020-01-01T13:00 is the value at 1pm, if subannual is given as 2020-01-01, it is understood as midnight that day, if it's January, it is understood as midnight on the first day of the month.
```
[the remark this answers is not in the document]
erikfilias: If I'm getting your point, a 3rd bullet could be added to clarify if the value is divided automatically or avoid this assumption.  I prefer the second option, so we could fall into many details that can be managed before. I think that we should upload data after the aggregation/disaggregation process and try to avoid any automatic math operation.  Any opinion?

nomenclature/definitions/subannual/README.md
```diff
@@ -43,6 +43,85 @@ with open(f'../subannual/months.yaml', 'r') as stream:
...
+> It is the total responsibility of each modelling team that must transform the data as part of the input/output processing workflow because the validate() function of nomenclature will validate how the timestamps are provided but it won't recognize the difference between UTC and local time as well as nor transform data from each other.
+
+#### Representatives timeslices:
+It refers to clustered hours and not consecutive that generates representatives timeslices. Some agreements are reflected in the following:
+
+- Using winter/summer time as representatives timeslices  :
+> To distinguish between different granularity levels of representative timeslices, It was proposed the following: `<Granularity>|<Name of timeslice>`. For example: `2 Season-2 Times|Summer-Day`.
+
+- Averaging values over the time span   :
+> A value is always the average for flow variables (i.e. It is the average between the subannual time and the subsequent one, where average is contingent on the lowest level of granularity - if you use 2020-01-01T13:00, it is hourly average, if you use 2020-01-01, it is daily average...).
```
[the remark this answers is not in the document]
erikfilias: It is related to power flow, inflows, etc (e.g. any fluids, I think).

erikfilias:
It is updating and improving in #66 .

PR #58, opened 2020-06-24
Addition of variables related to power plants from openTEPES

erikfilias:
Two commits:

- Moving variables related to CO2 emissions of power plants from power-plant.yaml to emissions.yaml
- Reorder and addition of variables in power-plant.yaml

### DIFF
--- a/nomenclature/definitions/variable/emissions/emissions.yaml
+++ b/nomenclature/definitions/variable/emissions/emissions.yaml
@@ -294,7 +294,7 @@ Emissions|CO2|Electricity:
    description: CO2 emmissions of power plants .
    unit: tons/MWh
 
-Emmissions|CO2|Electricity|Biomass|Traditionnal:
+Emissions|CO2|Electricity|Biomass|Traditionnal:
    description: CO2 emmissions of a typical Traditionnal Biomass power plant
                 biomass from local and pre-existing
                 resources mostly waste :domestic and agricultural waste, forest residues.
--- a/nomenclature/definitions/variable/emissions/emissions.yaml
+++ b/nomenclature/definitions/variable/emissions/emissions.yaml
@@ -295,34 +295,34 @@ Emissions|CO2|Electricity Generation:
    unit: tons/MWh
 
 Emissions|CO2|Electricity|Biomass|Traditionnal:
-   description: CO2 emmissions of a typical Traditionnal Biomass power plant
+   description: CO2 emissions of a typical Traditionnal Biomass power plant
                 biomass from local and pre-existing
-                resources mostly waste :domestic and agricultural waste, forest residues.
+                resources mostly waste :domestic and agricultural waste, forest residues
    unit: tons/MWh
 
 Emissions|CO2|Electricity|Biomass|New and imported :
-   description: CO2 emmissions of a typical new biomass power plant
-                biomass from dedicated energy crops and biomass imports.
+   description: CO2 emissions of a typical new biomass power plant
+                biomass from dedicated energy crops and biomass imports
    unit: tons/MWh
 
 Emissions|CO2|Electricity|Coal:
-   description: CO2 emmissions of a typical coal power plant
+   description: CO2 emissions of a typical coal power plant
    unit: tons/MWh
 
 Emissions|CO2|Electricity|Gas|OCGT:
-   description: CO2 emmissions of a typical Open Cycle Gas Turbine power plant
+   description: CO2 emissions of a typical Open Cycle Gas Turbine power plant
    unit: tons/MWh
 
 Emissions|CO2|Electricity|Gas|CCGT:
-   description: CO2 emmissions of a typical Combined Cycle Gas Turbine power plant
+   description: CO2 emissions of a typical Combined Cycle Gas Turbine power plant
    unit: tons/MWh
 
 Emissions|CO2|Electricity|Lignite:
-   description: CO2 emmissions of a typical Lignite power plant
+   description: CO2 emissions of a typical Lignite power plant
    unit: tons/MWh
 
 Emissions|CO2|Electricity|Nuclear:
-   description: CO2 emmissions of a typical nuclear power plant
+   description: CO2 emissions of a typical nuclear power plant
    unit: tons/MWh
 
 
--- a/nomenclature/definitions/variable/emissions/emissions.yaml
+++ b/nomenclature/definitions/variable/emissions/emissions.yaml
@@ -322,7 +322,7 @@ Emissions|CO2|Electricity|Lignite:
    unit: tons/MWh
 
 Emissions|CO2|Electricity|Nuclear:
-   description: CO2 emissions of a typical nuclear power plant
+   description: CO2 emmissions of a typical nuclear power plant; usually equal to 0 but variable necessary for consistancy reason
    unit: tons/MWh
 
 
--- a/nomenclature/definitions/variable/emissions/emissions.yaml
+++ b/nomenclature/definitions/variable/emissions/emissions.yaml
@@ -290,38 +290,34 @@ Emissions|CO2|AFOLU:
                 (IPCC category 3)
    unit: Mt CO2/yr
 
-Emissions|CO2|Electricity Generation:
-   description: CO2 emissions of power plants
-   unit: tons/MWh
-
-Emissions|CO2|Electricity|Biomass|Traditionnal:
+Emission Rate|CO2|Electricity|Biomass|Traditionnal:
    description: CO2 emissions of a typical Traditionnal Biomass power plant
                 biomass from local and pre-existing
                 resources mostly waste :domestic and agricultural waste, forest residues.
    unit: tons/MWh
 
-Emissions|CO2|Electricity|Biomass|New and imported :
+Emission Rate|CO2|Electricity|Biomass|New and imported :
    description: CO2 emissions of a typical new biomass power plant
                 biomass from dedicated energy crops and biomass imports.
    unit: tons/MWh
 
-Emissions|CO2|Electricity|Coal:
+Emission Rate|CO2|Electricity|Coal:
    description: CO2 emissions of a typical coal power plant
    unit: tons/MWh
 
-Emissions|CO2|Electricity|Gas|OCGT:
+Emission Rate|CO2|Electricity|Gas|OCGT:
    description: CO2 emissions of a typical Open Cycle Gas Turbine power plant
    unit: tons/MWh
 
-Emissions|CO2|Electricity|Gas|CCGT:
+Emission Rate|CO2|Electricity|Gas|CCGT:
    description: CO2 emissions of a typical Combined Cycle Gas Turbine power plant
    unit: tons/MWh
 
-Emissions|CO2|Electricity|Lignite:
+Emission Rate|CO2|Electricity|Lignite:
    description: CO2 emissions of a typical Lignite power plant
    unit: tons/MWh
 
-Emissions|CO2|Electricity|Nuclear:
+Emission Rate|CO2|Electricity|Nuclear:
    description: CO2 emissions of a typical nuclear power plant; usually equal to 0 but variable necessary for consistancy reason
    unit: tons/MWh
 
--- a/nomenclature/definitions/variable/emissions/emissions.yaml
+++ b/nomenclature/definitions/variable/emissions/emissions.yaml
@@ -321,7 +321,7 @@ Emissions|CO2|Electricity|Lignite:
    description: CO2 emmissions of a typical Lignite power plant
    unit: tons/MWh
 
-Emmissions|CO2|Electricity|Nuclear:
+Emissions|CO2|Electricity|Nuclear:
    description: CO2 emmissions of a typical nuclear power plant
    unit: tons/MWh
 
--- a/nomenclature/definitions/variable/emissions/emissions.yaml
+++ b/nomenclature/definitions/variable/emissions/emissions.yaml
@@ -300,7 +300,7 @@ Emissions|CO2|Electricity|Biomass|Traditionnal:
                 resources mostly waste :domestic and agricultural waste, forest residues
    unit: tons/MWh
 
-Emissions|CO2|Electricity|Biomass|New and imported :
+Emissions|CO2|Electricity|Biomass|New and imported:
    description: CO2 emissions of a typical new biomass power plant
                 biomass from dedicated energy crops and biomass imports
    unit: tons/MWh
--- a/nomenclature/definitions/variable/emissions/emissions.yaml
+++ b/nomenclature/definitions/variable/emissions/emissions.yaml
@@ -290,6 +290,42 @@ Emissions|CO2|AFOLU:
                 (IPCC category 3)
    unit: Mt CO2/yr
 
+Emmissions|CO2|Electricity:
+   description: CO2 emmissions of power plants .
+   unit: tons/MWh
+
+Emmissions|CO2|Electricity|Biomass|Traditionnal:
+   description: CO2 emmissions of a typical Traditionnal Biomass power plant
+                biomass from local and pre-existing
+                resources mostly waste :domestic and agricultural waste, forest residues.
+   unit: tons/MWh
+
+Emmissions|CO2|Electricity|Biomass|New and imported :
+   description: CO2 emmissions of a typical new biomass power plant
+                biomass from dedicated energy crops and biomass imports.
+   unit: tons/MWh
+
+Emmissions|CO2|Electricity|Coal:
+   description: CO2 emmissions of a typical coal power plant
+   unit: tons/MWh
+
+Emmissions|CO2|Electricity|Gas|OCGT:
+   description: CO2 emmissions of a typical Open Cycle Gas Turbine power plant
+   unit: tons/MWh
+
+Emmissions|CO2|Electricity|Gas|CCGT:
+   description: CO2 emmissions of a typical Combined Cycle Gas Turbine power plant
+   unit: tons/MWh
+
+Emmissions|CO2|Electricity|Lignite:
+   description: CO2 emmissions of a typical Lignite power plant
+   unit: tons/MWh
+
+Emmissions|CO2|Electricity|Nuclear:
+   description: CO2 emmissions of a typical nuclear power plant
+   unit: tons/MWh
+
+
 Emissions|CO2|Energy:
    description: CO2 emissions from energy use on supply and demand side (IPCC
                 category 1A, 1B)
--- a/nomenclature/definitions/variable/emissions/emissions.yaml
+++ b/nomenclature/definitions/variable/emissions/emissions.yaml
@@ -297,12 +297,12 @@ Emissions|CO2|Electricity Generation:
 Emissions|CO2|Electricity|Biomass|Traditionnal:
    description: CO2 emissions of a typical Traditionnal Biomass power plant
                 biomass from local and pre-existing
-                resources mostly waste :domestic and agricultural waste, forest residues
+                resources mostly waste :domestic and agricultural waste, forest residues.
    unit: tons/MWh
 
-Emissions|CO2|Electricity|Biomass|New and imported:
+Emissions|CO2|Electricity|Biomass|New and imported :
    description: CO2 emissions of a typical new biomass power plant
-                biomass from dedicated energy crops and biomass imports
+                biomass from dedicated energy crops and biomass imports.
    unit: tons/MWh
 
 Emissions|CO2|Electricity|Coal:
@@ -322,7 +322,7 @@ Emissions|CO2|Electricity|Lignite:
    unit: tons/MWh
 
 Emissions|CO2|Electricity|Nuclear:
-   description: CO2 emmissions of a typical nuclear power plant; usually equal to 0 but variable necessary for consistancy reason
+   description: CO2 emissions of a typical nuclear power plant; usually equal to 0 but variable necessary for consistancy reason
    unit: tons/MWh
 
 
--- a/nomenclature/definitions/variable/emissions/emissions.yaml
+++ b/nomenclature/definitions/variable/emissions/emissions.yaml
@@ -313,7 +313,7 @@ Emissions|CO2|Electricity|Gas|OCGT:
    description: CO2 emmissions of a typical Open Cycle Gas Turbine power plant
    unit: tons/MWh
 
-Emmissions|CO2|Electricity|Gas|CCGT:
+Emissions|CO2|Electricity|Gas|CCGT:
    description: CO2 emmissions of a typical Combined Cycle Gas Turbine power plant
    unit: tons/MWh
 
--- a/nomenclature/definitions/variable/emissions/emissions.yaml
+++ b/nomenclature/definitions/variable/emissions/emissions.yaml
@@ -290,7 +290,7 @@ Emissions|CO2|AFOLU:
                 (IPCC category 3)
    unit: Mt CO2/yr
 
-Emissions|CO2|Electricity:
+Emissions|CO2|Electricity Generation:
    description: CO2 emmissions of power plants .
    unit: tons/MWh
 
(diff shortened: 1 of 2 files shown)


nomenclature/definitions/variable/technology/power-plant.yaml
```diff
@@ -464,12 +630,45 @@ Inertia|Electricity|Hydro|Reservoir:
...
+   description: This is the upper bound that energy storage system could charge energy at time "t".
+   unit: GWh
+
+Initial Storage Level|Electricity|Energy Storage System:
+   description: Initial energy stored in the energy storage system unit at first load level. (ESS that includes hydro, battery, etc.)
+   unit: GWh
+
+Maximum Charge|Electricity|Energy Storage System:
+   description: Maximum charge when storing energy the energy storage system unit.
+   unit: MW
```
sebastianzwickl, line 651: Related to the conversation at https://github.com/openENTRANCE/nomenclature/issues/50#issuecomment-642560593 I suppose to add **Maximum Discharge**.

nomenclature/definitions/variable/emissions/emissions.yaml
```diff
@@ -290,6 +290,42 @@ Emissions|CO2|AFOLU:
                 (IPCC category 3)
    unit: Mt CO2/yr
 
+Emmissions|CO2|Electricity:
```
sandrinecharousset, line 293: Emmissions|CO2|Electricity:
```suggestion
Emissions|CO2|Electricity:
```

nomenclature/definitions/variable/emissions/emissions.yaml
```diff
@@ -290,6 +290,42 @@ Emissions|CO2|AFOLU:
...
+   description: CO2 emmissions of power plants .
+   unit: tons/MWh
+
+Emmissions|CO2|Electricity|Biomass|Traditionnal:
```
sandrinecharousset, line 297:
```suggestion
Emissions|CO2|Electricity|Biomass|Traditionnal:
```

nomenclature/definitions/variable/emissions/emissions.yaml
```diff
@@ -290,6 +290,42 @@ Emissions|CO2|AFOLU:
...
+   description: CO2 emmissions of a typical Traditionnal Biomass power plant
+                biomass from local and pre-existing
+                resources mostly waste :domestic and agricultural waste, forest residues.
+   unit: tons/MWh
+
+Emmissions|CO2|Electricity|Biomass|New and imported :
```
sandrinecharousset, line 303:
```suggestion
Emissions|CO2|Electricity|Biomass|New and imported :
```

nomenclature/definitions/variable/emissions/emissions.yaml
```diff
@@ -290,6 +290,42 @@ Emissions|CO2|AFOLU:
...
+   description: CO2 emmissions of a typical new biomass power plant
+                biomass from dedicated energy crops and biomass imports.
+   unit: tons/MWh
+
+Emmissions|CO2|Electricity|Coal:
```
sandrinecharousset, line 308:
```suggestion
Emissions|CO2|Electricity|Coal:
```

nomenclature/definitions/variable/emissions/emissions.yaml
```diff
@@ -290,6 +290,42 @@ Emissions|CO2|AFOLU:
...
+   description: CO2 emmissions of a typical coal power plant
+   unit: tons/MWh
+
+Emmissions|CO2|Electricity|Gas|OCGT:
```
sandrinecharousset, line 312:
```suggestion
Emissions|CO2|Electricity|Gas|OCGT:
```

nomenclature/definitions/variable/emissions/emissions.yaml
```diff
@@ -290,6 +290,42 @@ Emissions|CO2|AFOLU:
...
+   description: CO2 emmissions of a typical Open Cycle Gas Turbine power plant
+   unit: tons/MWh
+
+Emmissions|CO2|Electricity|Gas|CCGT:
```
sandrinecharousset, line 316:
```suggestion
Emissions|CO2|Electricity|Gas|CCGT:
```

nomenclature/definitions/variable/emissions/emissions.yaml
```diff
@@ -290,6 +290,42 @@ Emissions|CO2|AFOLU:
...
+   description: CO2 emmissions of a typical Combined Cycle Gas Turbine power plant
+   unit: tons/MWh
+
+Emmissions|CO2|Electricity|Lignite:
```
sandrinecharousset, line 320:
```suggestion
Emissions|CO2|Electricity|Lignite:
```

nomenclature/definitions/variable/emissions/emissions.yaml
```diff
@@ -290,6 +290,42 @@ Emissions|CO2|AFOLU:
...
+   description: CO2 emmissions of a typical Lignite power plant
+   unit: tons/MWh
+
+Emmissions|CO2|Electricity|Nuclear:
```
sandrinecharousset, line 324:
```suggestion
Emissions|CO2|Electricity|Nuclear:
```

nomenclature/definitions/variable/technology/power-plant.yaml
erikfilias: Ok, got it.

sebastianzwickl: Perfect, thank you @erikfilias

nomenclature/definitions/variable/emissions/emissions.yaml
```diff
@@ -290,6 +290,42 @@ Emissions|CO2|AFOLU:
                 (IPCC category 3)
    unit: Mt CO2/yr
 
+Emissions|CO2|Electricity:
```
danielhuppmann, line 293: I suggest to change this for clarification that this is not related to the consumption of electricity (and also implement for the other variables)
```suggestion
Emissions|CO2|Electricity Generation:
```

nomenclature/definitions/variable/emissions/emissions.yaml
```diff
@@ -290,6 +290,42 @@ Emissions|CO2|AFOLU:
...
+   description: CO2 emmissions of power plants .
```
danielhuppmann, line 294: typo, please correct throughout this document and remove `.`
```suggestion
   description: CO2 emissions of power plants
```

nomenclature/definitions/variable/emissions/emissions.yaml
```diff
@@ -290,6 +290,42 @@ Emissions|CO2|AFOLU:
...
+   unit: tons/MWh
+
+Emissions|CO2|Electricity|Biomass|Traditionnal:
+   description: CO2 emmissions of a typical Traditionnal Biomass power plant
+                biomass from local and pre-existing
+                resources mostly waste :domestic and agricultural waste, forest residues.
+   unit: tons/MWh
+
+Emissions|CO2|Electricity|Biomass|New and imported :
```
danielhuppmann, line 303: typo
```suggestion
Emissions|CO2|Electricity|Biomass|New and imported:
```

nomenclature/definitions/variable/emissions/emissions.yaml
```diff
@@ -290,6 +290,42 @@ Emissions|CO2|AFOLU:
                 (IPCC category 3)
    unit: Mt CO2/yr
 
+Emissions|CO2|Electricity:
+   description: CO2 emmissions of power plants .
+   unit: tons/MWh
```
danielhuppmann, line 295: I am skeptical to have this variable as "per unit of generated electricity", because other `Emissions|CO2|*` variables are reporting **total emissions** - so this could easily lead to confusion.

If you want to add a variable to describe operational characteristics, can we find an alternative name?

@sandrinecharousset what do you think?

nomenclature/definitions/variable/emissions/emissions.yaml
```diff
@@ -290,6 +290,42 @@ Emissions|CO2|AFOLU:
...
+
+Emissions|CO2|Electricity|Gas|CCGT:
+   description: CO2 emmissions of a typical Combined Cycle Gas Turbine power plant
+   unit: tons/MWh
+
+Emissions|CO2|Electricity|Lignite:
+   description: CO2 emmissions of a typical Lignite power plant
+   unit: tons/MWh
+
+Emissions|CO2|Electricity|Nuclear:
```
danielhuppmann, line 324: what are the CO2 emissions of a nuclear power plant during normal operations?

nomenclature/definitions/variable/technology/power-plant.yaml
```diff
@@ -1,4 +1,135 @@
-# description of thermal power plants
+# List of variables related to power plants
+# Variables related to the production cost of a power plant
+# Note that:
+# linear   term cost   [€/GWh] = Linear Var Cost [Mcal/GWh] * Price|Primary Energy [€/Mcal] + Operation and Maintenance Variable Cost [€/MWh] * 1e-3
+# constant term cost   [€/h]   = Constant Var Cost [Mcal/h] * Price|Primary Energy [€/Mcal]
+Constant Var Cost|Electricity|Coal:
```
danielhuppmann, line 6: suchest to rename that to
```suggestion
Operational Cost|Constant Term|Electricity Generation|Coal:
```

nomenclature/definitions/variable/technology/power-plant.yaml
```diff
@@ -1,4 +1,135 @@
...
+   description: Constant term (intercept) of the heat rate straight line. Cost of the power generation unit who belongs to the specified power
+                plant category. It refers to the unit commitment variable of each power plant per unit time.
+   unit: Mcal/hour
+
+Constant Var Cost|Electricity|Oil:
+   description: Constant term (intercept) of the heat rate straight line. Cost of the power generation unit who belongs to the specified power
+                plant category. It refers to the unit commitment variable of each power plant per unit time.
+   unit: Mcal/hour
+
+Linear Var Cost|Electricity|Coal:
```
danielhuppmann, line 26:
```suggestion
Operational Cost|Linear Term|Electricity Generation|Coal:
```

nomenclature/definitions/variable/technology/power-plant.yaml
```diff
@@ -1,4 +1,135 @@
...
+   description: Startup heat cost per power generation unit who belongs to the specified power
+                plant category.
+   unit: Mcal
+
+Start-Up Cost|Electricity|Oil:
+   description: Startup heat cost per power generation unit who belongs to the specified power
+                plant category.
+   unit: Mcal
+
+Variable Cost|Electricity|Biomass|Traditionnal:
```
danielhuppmann, line 105: Can these variables be removed as there is now a clear separation into linear and fixed components of the operational costs?

nomenclature/definitions/variable/technology/power-plant.yaml
```diff
@@ -44,7 +176,11 @@ Maximum Active power|Electricity|Gas|OCGT:
    description: Maximum active power of one typical Open Cycle Gas Turbine power plant
    unit: MW
 
-Maximum Active power|Electricity|Gas|Coal:
+Maximum Active power|Electricity|Gas|OCGT:
+   description: Maximum active power of one typical Combined Cycle Gas Turbine power plant
```
danielhuppmann, line 180: This is not the active power of the plant but normalized to the capacity (per MW)

nomenclature/definitions/variable/emissions/emissions.yaml
arght: For a power plant we can use emission rate

nomenclature/definitions/variable/emissions/emissions.yaml
arght: I understand that it is just written for consistency of dealing with all the conventional power plants

nomenclature/definitions/variable/emissions/emissions.yaml
sandrinecharousset: @danielhuppmann @arght I agree with the proposal of using emission rate instead of emission; I will include it in 'review mode' so that erik can accept changes easily (those are variables I did create in an older PR)

nomenclature/definitions/variable/emissions/emissions.yaml
sandrinecharousset: yes this is a consistency variable.... emissions of nuclear ar 0.....In my model, as in most UC models and it may be the same in comillas, nuclear is modelled as a thermal plant then it has the same list of variables (see issue #59 that I created about the lists of technologies : there are groups of technologies that share the same list of variables); Not sure what is the best to do there

nomenclature/definitions/variable/technology/power-plant.yaml
sandrinecharousset: yes; I think I did create those variables long ago, before erik added the more detailed ones from tepes.

nomenclature/definitions/variable/emissions/emissions.yaml
arght: should we propose the use of t for ton following the International System of Units?

nomenclature/definitions/variable/technology/power-plant.yaml
sandrinecharousset: this is the complex stuff here. When we model a generation mix that already exist, we can describe power plants one by one and each one has its own technical data (like its maximum active power....) and a name. When we are dealing with the future, what we do is that we describe a serie of  'typical' power plant (because we need to know what is te maximal active power of ONE single plant, as the model will include all the integer constraints....), usually one per technology per country (as countries may manage their plants differently, example: french nuclear plants are flexible and used intensively for providing ancillary services, while in most countries nuclear plants are operated only at nominal power, this can be seen in eg the minimum active power and the duration constraints) and we also describe the total capacity per technology per country. If we did not have the need of describing single units, then the Maximum Active Power variable would not be needed as it would be equal to the capacity.
To answer you comment, no this is not the normalized active power to the capacity (as this would be 1), it really something like 1400MW , 500MW...
Your comments means to me that the description should be improved.
@erikfilias any idea?

nomenclature/definitions/variable/emissions/emissions.yaml
sandrinecharousset:
```suggestion
Emission Rate|CO2|Electricity|Biomass|Traditionnal:
```

nomenclature/definitions/variable/emissions/emissions.yaml
sandrinecharousset:
```suggestion
Emission Rate|CO2|Electricity|Biomass|New and imported :
```

nomenclature/definitions/variable/emissions/emissions.yaml
sandrinecharousset:
```suggestion
Emission Rate|CO2|Electricity|Coal:
```

nomenclature/definitions/variable/emissions/emissions.yaml
sandrinecharousset:
```suggestion
Emission Rate|CO2|Electricity|Gas|OCGT:
```

nomenclature/definitions/variable/emissions/emissions.yaml
sandrinecharousset:
```suggestion
Emission Rate|CO2|Electricity|Gas|CCGT:
```

nomenclature/definitions/variable/emissions/emissions.yaml
sandrinecharousset:
```suggestion
Emission Rate|CO2|Electricity|Lignite:
```

nomenclature/definitions/variable/emissions/emissions.yaml
sandrinecharousset:
```suggestion
Emission Rate|CO2|Electricity|Nuclear:
```

nomenclature/definitions/variable/emissions/emissions.yaml
```diff
@@ -290,6 +290,42 @@ Emissions|CO2|AFOLU:
...
+Emmissions|CO2|Electricity|Gas|CCGT:
+   description: CO2 emmissions of a typical Combined Cycle Gas Turbine power plant
+   unit: tons/MWh
+
+Emmissions|CO2|Electricity|Lignite:
+   description: CO2 emmissions of a typical Lignite power plant
+   unit: tons/MWh
+
+Emmissions|CO2|Electricity|Nuclear:
+   description: CO2 emmissions of a typical nuclear power plant
```
sandrinecharousset, line 325:
```suggestion
   description: CO2 emmissions of a typical nuclear power plant; usually equal to 0 but variable necessary for consistancy reason
```

nomenclature/definitions/variable/technology/power-plant.yaml
sandrinecharousset: @erikfilias I suggest we delete all variable cost| variables

nomenclature/definitions/variable/emissions/emissions.yaml
erikfilias: I've just added this explanation to the PR. 
It's committed.

nomenclature/definitions/variable/technology/power-plant.yaml
erikfilias: I totally agree with you, @sandrinecharousset . 
In my point of view, we could make that maximum active power and capacity will be the same because in theory data related to power generation are the same in power and energy models. 

On the other hand, I wrote maximum active power because it is a standard to define the active and reactive power in power system models. Maybe, there will be other models related to AC load flow that use the data or to be included in our project in the future. This is the reason.

In addition, I'm curious about if we will compare the total capacity of power plants used in energy models with the values that we have in power models since the first one usually use a total amount, and the second one use values per power plant, node, region, or country. 

Then, can we conclude about mixing both terms and use only the term capacity?

nomenclature/definitions/variable/technology/power-plant.yaml
erikfilias: Ok, I got it.

nomenclature/definitions/variable/emissions/emissions.yaml
erikfilias: Ok, Thanks @sandrinecharousset

nomenclature/definitions/variable/emissions/emissions.yaml
erikfilias: Thanks, I got it.

nomenclature/definitions/variable/emissions/emissions.yaml
erikfilias: Thanks, I got it.

nomenclature/definitions/variable/emissions/emissions.yaml
erikfilias: Thanks, I got it.

nomenclature/definitions/variable/emissions/emissions.yaml
erikfilias: Thanks, I got it.

nomenclature/definitions/variable/emissions/emissions.yaml
erikfilias: Thanks, I got it.

nomenclature/definitions/variable/emissions/emissions.yaml
erikfilias: Thanks, I got it.

nomenclature/definitions/variable/emissions/emissions.yaml
sandrinecharousset: @arght yes I agree with t

nomenclature/definitions/variable/technology/power-plant.yaml
sandrinecharousset: @erikfilias for my model I need both the maximum active power of one plant and the capacity of the technology. Eg I could have for nuclear in france: capacity=50000MW and maximum active power = 1400MW. Thus the model will create the good number of individual units .
then the comparison you suggest in my case is perfectly equivalent :-)

nomenclature/definitions/variable/technology/power-plant.yaml
sandrinecharousset: @erikfilias @danielhuppmann I was a bit quick in answering. ACtually we need to keep the variable cost variable as it is not the same thing than the linear var cost. linear var cost is if I understood well a component of a formulae used by comillas to compute what I call the variable cost. In my model the cost for thermal unit is quadratic, then we have 3 inputs: quadratic cost, variable cost and fixed cost. the operation cost is then quadratic cost * active power * active power + variable cost * active power + fixed cost. As we do not have any public data available for quadratic cost and fixed cost I had only defined the variable cost. 
Maybe this is equivalent to Operation and Maintenance cost (as the unit is €/MWh) but then I would propose to define it somow differently, eg
Operation and Maintenance Cost|Linear term|Electricity|Gas
or
Operation and Maintenance Cost|Electricity|Gas|Linear term
@erikfilias I thus think that the variables like Operational Cost|Linear Term|Electricity Generation|Coal: should be named a bit differently as they are not really costs (the unit does not include €), and it is somehow ambiguous...

erikfilias: Yeah, I've just checked the differences. 
In openTEPES, we make a previous step in related to your variable cost.
E.g. we consider the following:
```
Operational Cost|Linear Term|Electricity Generation|...
Operation and Maintenance Cost|Electricity|...
Operational Cost|Constant Term|Electricity Generation|...
```
And, the next expressions are calculated:
```
**linear   term cost**   [€/GWh] = Operational Cost|Linear Term|Electricity Generation|... [Mcal/MWh] * Price|Primary Energy [€/Mcal] + Operation and Maintenance Cost|Electricity|... [€/MWh] * 1e-3
**constant term cost**   [€/h]   = Operational Cost|Constant Term|Electricity Generation|... [Mcal/h] * Price|Primary Energy [€/Mcal]
```

The **linear   term cost** is equivalent to your **variable cost**, and **constant term cost** to your fixed cost, I think. 

Then, I think we could keep these  variables because are not equal. And, I agree we should to rename it.
The  `Operational Cost|Linear Term|Electricity Generation|...` is defined as the linear term (slope) of the heat rate straight line, and `Operational Cost|Constant Term|Electricity Generation|...`, as the constant term (intercept) of the heat rate straight line.
Maybe, I could propose new names based on their definitions, @sandrinecharousset, do you agree?

nomenclature/definitions/variable/technology/power-plant.yaml
erikfilias: Really nice! :)
Then, I will keep the variables unless there is another contrary comment.

nomenclature/definitions/variable/technology/power-plant.yaml
sandrinecharousset: @erikfilias yes would be more clear and avoid confusions :-) 
We also use the same kind of formulaes in our models but depending on the datasource, usually what you can find is the variable cost only..... Then in long-term models we use that one. (and in short term operational models we use the ones computed out of the formulas, but in that case we do have all the technical data)

arght: I suggest to express startup and shutdown costs in MEUR instead of Mcal.

danielhuppmann: what is "MEUR"? if million, "million EUR", "mEUR" might be more intuitive...

arght: I suggest to keep M to be compatible with the International System of Units

nomenclature/definitions/variable/emissions/emissions.yaml
erikfilias: I made it locally some days ago because the commit suggestion was outdated

erikfilias:
I rebase this PR, and made one commit related to the use of "EUR".

danielhuppmann:
merging as is and then implement a clean-up in a separate PR later

PR #62, opened 2020-06-29
Fix validation fails on directional data 

### LINKED ISSUE #60: Validation fails on directional data
The current `validate()`function fails on any data using directional region definitions (i.e., `Region A>Region B`).

Reported by @sandrinecharousset

sebastianzwickl:
Update the validate function to validate directional data (i.e., RegionA>RegionB). Should fix #60


nomenclature/__init__.py
```diff
@@ -109,8 +107,12 @@ def validate(df):
     msg = 'The following {} are not defined in the nomenclature:\n    {}'
     for col, codelist, ext in cols:
         invalid = [c for c in df.data[col].unique() if c not in codelist]
+        for i in reversed(invalid):
```
danielhuppmann, line 110: please add a comment here to explain what this section is doing.

also, why is this `reversed`? the following should also just work
```suggestion
        for i in invalid:
```

nomenclature/__init__.py
```diff
@@ -93,10 +93,8 @@ def validate(df):
         Return `True` if all column entries in `df` are valid
         or `False` otherwise
     """
-    if not isinstance(df, IamDataFrame):
-        df = IamDataFrame(df)
+    df = IamDataFrame(df)
     success = True
-
```
danielhuppmann, line 99: I'd prefer to keep this empty line to better structure the code...

nomenclature/__init__.py
sebastianzwickl: I had minor difficulties with the remove command and iteration through the list. From my point of view, it is a _classic_ way to iterate through the list from behind and remove all valid entries from the invalid list. However, I will add more comments for clarification.

nomenclature/__init__.py
sebastianzwickl: Yes, that is true.

danielhuppmann:
closing and re-opening PR to start @stickler-ci

PR #64, opened 2020-07-24
Restructure the yaml file format to separate variable types from fuels

danielhuppmann:
This PR is a first step towards implementing the proposal in #29, namely to separate the list of fuels from the variable definitions. This will make the yaml files more readable (because shorter) and reduce the potential for errors (because avoiding unnecessary repetition).

The key change: the variables for electricity generation by fuel types (including CCS subcategories) are generically given by `Secondary Energy|Electricity|<Fuel>` (rather than having to define the full list in `energy-secondary.yaml`) and the `<Fuel>` tag will be replaced by the list given in `fuel_types.yaml` when importing the Python package (so that the validation continues to work as expected). A test is added to make sure that the correct variable and description are available via the Python package.

This PR only changed this one instance (electricity generation) to elicit feedback - other variables can then be simplified in a similar manner after a first round of discussion, e.g., replacing numerous occurrences of `Maximum Active Power|Electricity|...` by just one variable `Maximum Active Power|Electricity|<Fuel>` in `power-plant.yaml`.

PR #67, opened 2020-08-07
Refactoring and simplifying of variables related to power plant specifications

danielhuppmann:
This PR implements the refactoring of the fuel-type list (implemented for electricity at the secondary-energy level, see #64) for the variables representing characteristics of power plants.

A few clarifications are needed to finalize this:

 - Are we ok that there are multiple variables for operational cost
   - `Operational Cost|Constant Term|...`
   - `Operational Cost|Linear Term|...`
   - `Operation and Maintenance Cost|...`
  What are the differences?
 - Are we fine with the variables `Rate Frequency Containment Reserve` and `Rate Automatic Frequency Containment Reserve`, or should we drop the `Rate`?
 - Profile is defined in energy terms for hydro-reservoir inflow but as percentage (of what?) for other renewables... We should harmonize this before starting to use this.
 - What is meant with "Energy Storage System" - battery? compressed air?

### DIFF
--- a/nomenclature/definitions/variable/technology/power-plant.yaml
+++ b/nomenclature/definitions/variable/technology/power-plant.yaml
@@ -3,638 +3,153 @@
 # Note that:
 # linear   term cost   [EUR/GWh] = Linear Var Cost [Mcal/MWh] * Price|Primary Energy [EUR/Mcal] + Operation and Maintenance Variable Cost [EUR/MWh] * 1e-3
 # constant term cost   [EUR/h]   = Constant Var Cost [Mcal/h] * Price|Primary Energy [EUR/Mcal]
-Operational Cost|Constant Term|Electricity Generation|Coal:
-   description: Constant term (intercept) of the heat rate straight line. Cost of the power generation unit who belongs to the specified power
-                plant category. It refers to the unit commitment variable of each power plant per unit time.
-   unit: Mcal/hour
 
-Operational Cost|Constant Term|Electricity Generation|Gas:
-   description: Constant term (intercept) of the heat rate straight line. Cost of the power generation unit who belongs to the specified power
-                plant category. It refers to the unit commitment variable of each power plant per unit time.
-   unit: Mcal/hour
-
-Operational Cost|Constant Term|Electricity Generation|Nuclear:
-   description: Constant term (intercept) of the heat rate straight line. Cost of the power generation unit who belongs to the specified power
-                plant category. It refers to the unit commitment variable of each power plant per unit time.
-   unit: Mcal/hour
-
-Operational Cost|Constant Term|Electricity Generation|Oil:
-   description: Constant term (intercept) of the heat rate straight line. Cost of the power generation unit who belongs to the specified power
-                plant category. It refers to the unit commitment variable of each power plant per unit time.
-   unit: Mcal/hour
-
-Operational Cost|Linear Term|Electricity Generation|Coal:
-   description: Linear term (slope) of the heat rate straight line. Cost of the power generation unit who belongs to the specified power
-                plant category. It refers to the production cost of each energy unit (GWh, MWh,etc.).
-   unit: Mcal/MWh
-
-Operational Cost|Linear Term|Electricity Generation|Gas:
-   description: Linear term (slope) of the heat rate straight line. Cost of the power generation unit who belongs to the specified power
-                plant category. It refers to the production cost of each energy unit (GWh, MWh,etc.).
-   unit: Mcal/MWh
-
-Operational Cost|Linear Term|Electricity Generation|Nuclear:
-   description: Linear term (slope) of the heat rate straight line. Cost of the power generation unit who belongs to the specified power
-                plant category. It refers to the production cost of each energy unit (GWh, MWh,etc.).
-   unit: Mcal/MWh
-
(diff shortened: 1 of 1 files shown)


nomenclature/definitions/variable/technology/power-plant.yaml
```diff
@@ -3,638 +3,153 @@
...
-   description: Percentage of the Maximum Active Power that can be devoted to Frequency Containment Reserve
-                for a typical "simple Hydro reservoir " unit
-   unit: "%"
-
-Rate Automatic Frequency Restoration Reserve|Electricity|Hydro|Reservoir:
-   description: Percentage of the Maximum Active Power that can be devoted to Automatic Frequency Restoration Reserve
-                for a typical "simple Hydro reservoir " unit
+# specific variables for run-of-river hydro power plants
+Load Factor|Electricity|Hydro|Run of River|Profile:
+   description: Load of hourly loadfactors for "Run of River"
```
tburandt, line 152: As the name of the variable was changed, I would also change the description
```suggestion
   description: Load of hourly load factors for "Run of River"
```

nomenclature/definitions/variable/technology/power-plant.yaml
```diff
@@ -675,94 +190,16 @@ Stored Energy Inventory|Electricity|Energy Storage System:
...
-   description: Number of units of kind Photovoltaic power plant
-   unit:
-
-Maximum Active power|Electricity|Solar:
-   description: Maximum active power of one Photovoltaic power plant
-   unit: MW
-
-LoadFactor|Electricity|Solar|Profile:
+Load Factor|Electricity|Solar|Profile:
    description: Timeserie of hourly loadfactors for Photovoltaic power plants
```
tburandt, line 194:
```suggestion
   description: Timeseries of hourly load factors for Photovoltaic power plants
```

nomenclature/definitions/variable/technology/power-plant.yaml
```diff
@@ -675,94 +190,16 @@ Stored Energy Inventory|Electricity|Energy Storage System:
...
-   description: Number of units of kind Wind Onshore power plant
-   unit:
-
-Maximum Active power|Electricity|Wind|OnShore:
-   description: Maximum active power of one Wind Onshore power plant
-   unit: MW
-
-LoadFactor|Electricity|Wind|OnShore|Profile:
+Load Factor|Electricity|Wind|Onshore|Profile:
    description: Timeserie of hourly loadfactors for Onshore Windpower
```
tburandt, line 199:
```suggestion
   description: Timeseries of hourly load factors for Wind Onshore power plants
```

nomenclature/definitions/variable/technology/power-plant.yaml
```diff
@@ -675,94 +190,16 @@ Stored Energy Inventory|Electricity|Energy Storage System:
...
-   description: Number of units of kind Wind OffShore power plant
-   unit:
-
-Maximum Active power|Electricity|Wind|OffShore:
-   description: Maximum active power of one Wind OffShore power plant
-   unit: MW
-
-LoadFactor|Electricity|Wind|OffShore|Profile:
+Load Factor|Electricity|Wind|OffShore|Profile:
    description: Timeserie of hourly loadfactors for Offshore Windpower
```
tburandt, line 204:
```suggestion
   description: Timeseries of hourly load factors for Wind Offshore power plants
```

danielhuppmann:
thanks @tburandt, I accepted the suggestions (but they will probably change again depending on the conclusion about load factor convention)

arght:
- Are we fine with the variables Rate Frequency Containment Reserve and Rate Automatic Frequency Containment Reserve, or should we drop the Rate?

Usually the values that are included here depend on the demand, variable RES and largest generating unit. From this point of view I would rather prefer to specify these values in MW, not as a rate, which I assume would be in p.u. or %

arght:
•	Profile is defined in energy terms for hydro-reservoir inflow but as percentage (of what?) for other renewables... We should harmonize this before starting to use this.

Natural hydro inflows are usually expressed in m3/s or GWh for each reservoir. I understand that profile for variable RES can be expressed as p.u. of their installed capacity. Alternatively I suggest to express it directly as MW.

arght:
•	What is meant with "Energy Storage System" - battery? compressed air?
ESS is any system able to store and generate energy (pumped-storage hydro, battery, compressed air energy system, etc.)

sandrinecharousset:
Regarding the variables Rate Frequency Containment Reserve and Rate Automatic Frequency Containment Reserve, it is important to keep sth like rate or another word meaning it is a percentage OR these values could be expressed as maximum Frequency Containment Reserve (which is equAl to maximum active power * rate frequency....)
I am indifferent to this

sandrinecharousset:
All profiles for res could be expressed in MW or % of a maximal active power, as you prefer
This mean we would have a fixed max power AND a max power profile (keeping the |profile as for the inflows would be enough to distinguish)
Regarding inflows they should be in MWh not MW (equivalent if hourly but usually available data are daily or per 6h... but never saw hourly data...)

sandrinecharousset (approved):

danielhuppmann:
closing in favor of #90 #93 #96

PR #69, opened 2020-09-25
add yaml file with 5 Norwegian price regions

SarahYS:
Hi, 

EMPS provides its Norwegian output data according to bidding zones/price areas. We would need this to be included in the nomencalture


danielhuppmann (changes requested):
Thank you @SarahYS for the suggestion and PR! This looks good in principle, but I would suggest a few changes:

 - implementation-wise, please add it to the end of the `subcountries.yaml` file (to avoid having too many files)
 - I'd suggest to give a more intuitive name, e.g. `Norway|Bergen` rather than `NOP5`. Imagine a user seeing that abbreviation in the openENTRANCE Scenario Explorer - he/she will be confused what this means... If the price-region-numbering is taken from an official source, please add the source as a note.
 - Please take a look at the additional attributes (mapping to NUTS, note) given for other [sub-country regions](https://github.com/openENTRANCE/nomenclature/blob/9f84df6b1a1f3a91f9e299348ae91dc9455ccdca/nomenclature/definitions/region/subcountries.yaml#L20) and add similar helpful information.

SarahYS:
Hi Daniel, 

thank for the input. I have added the names to the existing subcountries.yaml file and changed the name. Regarding the mapping to NUTS, we are still working on it - thinking of opening a separate issue on that, since there is more discussion needed. Please let me know if there is something else I can improve. 

I have also added a missing "r" in \nomenclature\nomenclature\definitions\variable\technology\power-plant.yaml (Sebastian spotted it). 

Best,
Sarah

danielhuppmann (approved):
Thank you for the revisions. I created #71 as a reminder to still add NUTS-x mapping (if possible).

PR #70, opened 2020-09-30
Addition of Spanish super-grids from ehighway2050

erikfilias:
This PR adds Spanish regions that were defined and used in the ehighway2050 public datasets


danielhuppmann (approved):
thank you @erikfilias for this PR, looks good!

PR #75, opened 2020-11-09
Modify Norwegian price regions

SarahYS:
Hi, 

we have decided to increase the region resolution within Norway for our case study, so I have change/added the names for the new classification. 

Thank you,
Sarah

### DIFF
--- a/nomenclature/definitions/region/subcountries.yaml
+++ b/nomenclature/definitions/region/subcountries.yaml
@@ -305,80 +305,79 @@ United Kingdom|North Sea:
 # List of subregions in Norway
 # based on the regional electricity grid and key watersheds for hydropower generation
 Norway|Ostland:
-    definition: 'virtual' region holding hydropower system in Ostland (including plants connected by same watercourse)
+    definition: A 'virtual' region holding hydropower system in Ostland (including plants connected by same watercourse)
     country: Norway
     notes: This region was defined in the HydroCen dataset based on the Norwegian hydropower system and electricity price regions
     HydroCen: N1
     source: https://energifaktanorge.no/en/norsk-energiforsyning/kraftmarkedet/
 
 Norway|Sorost:
-    definition: 'virtual' region holding hydropower system in Olso and Viken (including plants connected by same watercourse
+    definition: A 'virtual' region holding hydropower system in Olso and Viken (including plants connected by same watercourse
     country: Norway
     notes: This region was defined in the HydroCen dataset based on the Norwegian hydropower system and electricity price regions
     HydroCen: N2
     source: https://energifaktanorge.no/en/norsk-energiforsyning/kraftmarkedet/
 
 Norway|Hallingdal:
-    definition: 'virtual' region holding hydropower system in Hallingdal (including plants connected by same watercourse)
+    definition: A 'virtual' region holding hydropower system in Hallingdal (including plants connected by same watercourse)
     country: Norway
     notes: This region was defined in the HydroCen dataset based on the Norwegian hydropower system and electricity price regions
     HydroCen: N3
     source: https://energifaktanorge.no/en/norsk-energiforsyning/kraftmarkedet/
 
 Norway|Telemark:
-    definition: 'virtual' region holding hydropower system in Telemark (including plants connected by same watercourse)
+    definition: A 'virtual' region holding hydropower system in Telemark (including plants connected by same watercourse)
     country: Norway
     notes: This region was defined in the HydroCen dataset based on the Norwegian hydropower system and electricity price regions
     HydroCen: N4
     source: https://energifaktanorge.no/en/norsk-energiforsyning/kraftmarkedet/
 
 Norway|Sorland:
-   definition: 'virtual' region holding hydropower system in Sørland (including plants connected by same watercourse)
+   definition: A 'virtual' region holding hydropower system in Sørland (including plants connected by same watercourse)
    country: Norway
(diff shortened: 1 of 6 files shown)


nomenclature/definitions/region/subcountries.yaml
```diff
@@ -302,28 +302,58 @@ United Kingdom|North Sea:
     ehighway2050: 106_NS+107_NS+108_NS+109_NS
     plan4res: NSUK
     
-# List of electricity price regions in Norway
-Norway|Oslo:
-   definition: this is the electricity price region 1 in Norway 
+# List of subregions in Norway, designed to best decribe the hydropower system in Norway
+Norway|OSTLAND:
+   definition: this is region N1 in the HydroCen dataset, these regions are based on the norwegian electricicty regions but with finer granularity (and some changes)
```
danielhuppmann, line 307: Not necessary to write "and some changes" to the definition, because future users won't know that there was an earlier version.

Please follow the convention used for the other sub-country regions, see the following suggestion
```suggestion
   definition: <please give a self-explaining definition, using NUTS-x codes if possible
   note: This region was defined in the HydroCen dataset based on the Norwegian electricity regions
   hydrocen: N1
```

nomenclature/definitions/region/subcountries.yaml
```diff
@@ -302,28 +302,58 @@ United Kingdom|North Sea:
     ehighway2050: 106_NS+107_NS+108_NS+109_NS
     plan4res: NSUK
     
-# List of electricity price regions in Norway
-Norway|Oslo:
-   definition: this is the electricity price region 1 in Norway 
+# List of subregions in Norway, designed to best decribe the hydropower system in Norway
```
danielhuppmann, line 305: Not sure if mentioning hydropower instead of electricity is helpful...

danielhuppmann (changes requested):
Thanks for the PR. Added a few notes. Also, please confirm that the previously defined regions should be removed.

nomenclature/definitions/region/subcountries.yaml
```diff
@@ -302,28 +302,58 @@ United Kingdom|North Sea:
...
-Norway|Kristiansand:
-   definition: this is the electricity price region 2 in Norway 
+Norway|SOROST:
+   definition: this is region N2 in the HydroCen dataset, these regions are based on the norwegian electricicty regions but with finer granularity (and some changes) 
    country: Norway
    source: https://energifaktanorge.no/en/norsk-energiforsyning/kraftmarkedet/
 
-Norway|Trondheim:
-   definition: this is the electricity price region 3 in Norway 
+Norway|HALLNGDAL:
```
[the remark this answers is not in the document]
SarahYS: Yes... Just doublechecked HALLINGDAL

nomenclature/definitions/region/subcountries.yaml
SarahYS: Ok, good point. I was explained that the regions are defined in a way to best respect the connections between different hydro power plants along a watercourse, hence not necessarily coinciding with electricity price regions are nuts regions. Maybe we could write "List of Norwegian sub-regions, designed to enable detailed modelling of the hydro power production system"? Do you think that would be better?

nomenclature/definitions/region/subcountries.yaml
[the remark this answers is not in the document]
danielhuppmann: Then there is an "I" missing in the region name... Also, are you sure that you want to have all regions written in all-caps?

nomenclature/definitions/region/subcountries.yaml
danielhuppmann: Please implement this change for all regions, not just the one region where I proposed it

nomenclature/definitions/region/subcountries.yaml
danielhuppmann: Maybe the following...
```suggestion
# List of subregions in Norway
# based on the regional electricity grid and key watersheds for hydropower generation
```

nomenclature/definitions/region/subcountries.yaml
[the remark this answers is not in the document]
SarahYS: Yes, the naming in all the HydroCen datasets are all capital... I thought then I might best stick to it?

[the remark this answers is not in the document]
danielhuppmann: For consistency with other region names and for better readability, I would use "normal" spelling rather than all-caps. If one wants to write a script to automatically convert from HydroCen to openENRTANCE in a workflow, a user can easily generate a mapping like `{'HALLINGDAL': 'Norway|Hallingdal'}`.

SarahYS:
Now I have changed the descriptions - sorry for all the million commits

nomenclature/definitions/region/subcountries.yaml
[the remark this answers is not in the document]
SarahYS: I have changed that now (if I succeeded in providing the right file.. )

nomenclature/definitions/region/subcountries.yaml
SarahYS: ok should be done

nomenclature/definitions/region/subcountries.yaml
SarahYS: I adjusted it

danielhuppmann (approved):
Thanks for this PR @SarahYS!

danielhuppmann:
For completeness, the regions removed by this PR will be deleted in the openENTRANCE Scenario Explorer instance.

PR #86, opened 2020-11-23
Updated electricity operation according to new fuel definition

sandrinecharousset:
This PR should follow #67 as it does the same on electicity-operation.yaml


nomenclature/definitions/variable/technology/electrity-operation.yaml
```diff
@@ -15,93 +15,22 @@
...
-   description: Active power of Combined Cycle Gas Turbine power plants
+Active Power|Electricity|<Fuel>:
+   description: Active power of <this fuel>
    unit: MWh
 
-Active Power|Electricity|Coal:
-   description: Active power of coal power plants
-   unit: MWh
+# Active Pumping Power per technologies
+# This is the Power consumed for pumping; it usually is a negative figure
```
danielhuppmann, line 23: Why is "active pumping power" a negative figure? I find that quite unintuitive...

danielhuppmann (commented):
Thank you @sandrinecharousset for implementing this simplification!

There is one clarification about why pumping-power should be negative before merging.

nomenclature/definitions/variable/technology/electrity-operation.yaml
sandrinecharousset: active pumping power is the power used for pumping water from downstream to upstream. The rule used is that generated power is positive and used power is negative. You may also see it as power taken from the grid instead of power sent to the grid. It is necessary to have a clear convention then that power taken from the grid is negative while power sent to the grid is positive. This convention is used in all electricity models.

danielhuppmann: I understand that you need to treat this value as negative in the energy-balance constraint, but I fear that if modellers just see the name "Active Pumping Power" (with a definition of consumption) and negative values, this can lead to confusion.

So I would rather change the definition to "electricity withdrawn from the grid for pumping in hydro storage reservoirs" understood as positive values, and then make sure that when this is translated from/to the modelling framework, it is multiplied by -1.

sandrinecharousset: I fear exactly the opposite, that some modelers may be tempted to sum all ActivePower* and if the pumping is positive than get to a false balance.
@erikfilias any opinion?

danielhuppmann: But the variable is called `Active Pumping Power|...`, which should be a sufficient warning to not aggregate it with `Active Power|...`.

erikfilias: I agree with @sandrinecharousset. We also used two variable generator output (discharge if an ESS) and consumption (charge if an ESS) in openTEPES. It is a suit way, I think. And, we should adopt it something like this, here. It could be: `Active Powe|Generation|...` and `Active Powe|Consumption|...`, in general for PSH, ESS, and batteries.

P.D. more details about this could be found in openTEPES' formulation. ([here](https://pascua.iit.comillas.edu/aramos/openTEPES/MathematicalFormulation.html)).

sandrinecharousset: @erikfilias I think your proposal is a very good idea and could solve any potential misunderstandings
If @danielhuppmann agrees I propose to change all Active Power variables into Generation|Active Power and Active Pumping power into Pumping Consumption|Active power or sth like that
Any advise welcome

danielhuppmann: Well, if it is renamed to `Active Power`, then I would have less worries about making this (expected to be) a negative value for pumping in storage plants. What struck me as counter-intertuitive was to specifically name it for something that is negative from the energy-balance point of view, and then also assign negative values.

So fine by me to simply go with `Active Power|Electricity|Hydro|Pumped Storage` (which already exists) and add a note that this is positive when water is discharged and negative during pumping.

erikfilias: We can then have only one variable that can be positive or negative to describe this behavior (`Active Power|Electricity|...`). Or to create two positive variables for charging and discharging. I prefer the second one. @sandrinecharousset, @arght, @sebastianzwickl, @tperger, any opinion?

sandrinecharousset: > I understand that you need to treat this value as negative in the energy-balance constraint, but I fear that if modellers just see the name "Active Pumping Power" (with a definition of consumption) and negative values, this can lead to confusion.
> 
> So I would rather change the definition to "electricity withdrawn from the grid for pumping in hydro storage reservoirs" understood as positive values, and then make sure that when this is translated from/to the modelling framework, it is multiplied by -1.

=> Pumped Storage is used to define a category of hydro power, different from 'reservoir' , not only the energy used for pumping, so I do not like very much this idea..... For hydro, generally the different kind of units are decomposed into 'run-of-river', 'pumped storage', which is units with small reservoirs and pumping capacities, and 'reservoir' which is units with huge reservoir (mountain lakes) and usually no pumping capacity (but may have in some cases)
For all those units we will have different outputs: power generated, and power consumed for pumping. The first one (the Active power) is positive, while the second is power consumed so it is negative.

sandrinecharousset:
All these changes are already in the PR by @erikfilias so I suggest to remove this one

PR #73, opened 2020-10-30
Variables for Non-Power Sectors in GENeSYS-MOD

tburandt:
This pull-requests adds the variables needed for sending non-power sector related data from GENeSYS-MOD to the openEntrance-Database. The added variables suggested in this pull-requests are based on Deliverable 5.2 "Definition of the interface between the models in the suite and the Common Database" of openEntrance.

It is also related to #7.

PR #88, opened 2021-01-19
Changed fixed cost

tburandt:
Changed fixed cost for industrial heat to match the new definition of variable and capital costs, as defined in #73.


danielhuppmann (approved):
Thanks for catching this @tburandt

PR #93, opened 2021-02-03
Harmonize variable name for run-of-river-hydro load factor

danielhuppmann:
This PR adds a space in the term "Load Factor" in line with the recommendation for variable naming. It also removes the "Profile" part from the variable name following the same logic as #92.


sandrinecharousset (approved):
you are completely right, profile is useless here, will be clear from the timestamps

PR #99, opened 2021-02-12
Change unit of `Mean Outage Duration` to time domain

### LINKED ISSUE #98: Mean Outage Duration - share or time?
While restructuring the variables related to power plants, I noticed that the variable "Mean Outage Duration" is given as % rather than a time (hour, seconds). Is there a particular reason? Or if this should be given as a percentage, then what is the base from the share is computed?

PS. This question also applies (to some extend) to "Forced Outage Rate" and "Planned Outage Rate"...

https://github.com/openENTRANCE/nomenclature/blob/cefbaf0bdd8afaf1b574a9dc69f6c48881b1e793/nomenclature/definitions/variable/technology/power-plant.yaml#L65

danielhuppmann:
This PR implements the change agreed in #98. To provide some flexibility, the units allowed are days (for large thermal plants) and hours (for smaller units where outages are more quickly resolved).

closes #98


danielhuppmann:
closing and opening to restart the tests

PR #103, opened 2021-02-24
Including regions from ehighway2050

erikfilias:
This PR adds the regions that were defined and used in the ehighway2050 public datasets but not included in Nomenclature.

### DIFF
--- a/nomenclature/definitions/region/subcountries.yaml
+++ b/nomenclature/definitions/region/subcountries.yaml
@@ -608,6 +608,44 @@ Sweden|Subgrid 89:
   note: This region was defined and used in the ehighway2050 public datasets
   ehighway2050: 88_SE
 
+# ehighway clusters United Kingdom
+
+United Kingdom|Subgrid 90:
+  definition: UKH11 + UKH12 + UKH14 + UKH15 + UKH16 + UKH17 + UKH21 + UKH23 + UKH24 + UKH25 + UKH31 + UKH32 + UKH34 + UKH35 + UKH36 + UKH37 + UKI31 + UKI32 + UKI33 + UKI34 + UKI41 + UKI42 + UKI43 + UKI44 + UKI45 + UKI51 + UKI52 + UKI53 + UKI54 + UKI61 + UKI62 + UKI63 + UKI71 + UKI72 + UKI73 + UKI74 + UKI75 + UKJ11 + UKJ12 + UKJ13 + UKJ14 + UKJ21 + UKJ22 + UKJ25 + UKJ26 + UKJ27 + UKJ28 + UKJ31 + UKJ32 + UKJ34 + UKJ35 + UKJ36 + UKJ37 + UKJ41 + UKJ43 + UKJ44 + UKJ45 + UKJ46
+  country: United Kingdom
+  note: This region was defined and used in the ehighway2050 public datasets
+  ehighway2050: 90_UK
+
+United Kingdom|Subgrid 91:
+  definition: UKK11 + UKK12 + UKK13 + UKK14 + UKK15 + UKK23 + UKK24 + UKK25 + UKK30 + UKK41 + UKK42 + UKK43
+  country: United Kingdom
+  note: This region was defined and used in the ehighway2050 public datasets
+  ehighway2050: 91_UK
+
+United Kingdom|Subgrid 92:
+  definition: UKD33 + UKD34 + UKD35 + UKD36 + UKD37 + UKD41 + UKD42 + UKD44 + UKD45 + UKD46 + UKD47 + UKD61 + UKD62 + UKD63 + UKD71 + UKD72 + UKD73 + UKD74 + UKE13 + UKE31 + UKE32 + UKE41 + UKE42 + UKE44 + UKE45 + UKF11 + UKF12 + UKF13 + UKF14 + UKF15 + UKF16 + UKF21 + UKF22 + UKF24 + UKF25 + UKF30 + UKG11 + UKG12 + UKG13 + UKG21 + UKG22 + UKG23 + UKG24 + UKG31 + UKG32 + UKG33 + UKG36 + UKG37 + UKG38 + UKG39 + UKL11 + UKL12 + UKL13 + UKL14 + UKL15 + UKL16 + UKL17 + UKL18 + UKL21 + UKL22 + UKL23 + UKL24
+  country: United Kingdom
+  note: This region was defined and used in the ehighway2050 public datasets
+  ehighway2050: 92_UK
+
+United Kingdom|Subgrid 93:
+  definition: UKC11 + UKC12 + UKC13 + UKC14 + UKC21 + UKC22 + UKC23 + UKD11 + UKD12 + UKE11 + UKE12 + UKE21 + UKE22 + UKM73 + UKM75 + UKM78 + UKM91 + UKM92 + UKM93 + UKM94 + UKM95
+  country: United Kingdom
+  note: This region was defined and used in the ehighway2050 public datasets
+  ehighway2050: 93_UK
+
+United Kingdom|Subgrid 94:
+  definition: UKM50 + UKM61 + UKM62 + UKM63 + UKM64 + UKM65 + UKM66 + UKM71 + UKM72 + UKM76 + UKM77 + UKM81 + UKM82 + UKM83 + UKM84
+  country: United Kingdom
+  note: This region was defined and used in the ehighway2050 public datasets
+  ehighway2050: 94_UK
+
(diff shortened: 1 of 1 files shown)


nomenclature/definitions/region/subcountries.yaml
```diff
@@ -275,33 +719,81 @@ Italy|Subgrid south:
 
 # ehighway north sea clusters
 Belgium|North Sea:
-    definition: this 'virtual' region holds the offshore windpower of belgium in norh sea
+    definition: This 'virtual' region holds the offshore windpower of belgium in norh sea
```
danielhuppmann, line 722: There is a typo here (and in all similar subgrid definitions, and proposing a slight rewording
```suggestion
    definition: This 'virtual' region represents offshore wind generation of Belgium in the North Sea
```

nomenclature/definitions/region/subcountries.yaml
```diff
@@ -187,73 +394,310 @@ Italy|Subgrid 98:
...
+Ireland|Subgrid 96:
+  definition: IE041 + IE042 + IE051 + IE052 + IE053 + IE061 + IE062 + IE063
+  country: United Kingdom
+  note: This region was defined and used in the ehighway2050 public datasets
+  ehighway2050: 96_IE
+
+# ehighway clusters
+
+Russia|Subgrid 100:
+  definition: Region without NUTS3 but defined in ehighway2050
```
danielhuppmann, line 660: Better to use some positive description (maybe "Russia west of the Urals"?) rather than stating how it is not possible to describe it. Also, the reference to ehighways2050 is already included as a not, so no need to repeat.

danielhuppmann (changes requested):
Thanks @erikfilias - two suggestions to improve the definitions (each applies to several regions). I didn't look at the NUTS3-mappings, but it looks like it makes sense.

sandrinecharousset (approved):

nomenclature/definitions/region/subcountries.yaml
erikfilias: Ok, thanks. Got it.

erikfilias: Please, check it.

nomenclature/definitions/region/subcountries.yaml
```diff
@@ -187,73 +394,310 @@ Italy|Subgrid 98:
...
+
+Russia|Subgrid 100:
+  definition: Russia West of the Urals
+  country: Russia
+  note: This region was defined and used in the ehighway2050 public datasets
+  ehighway2050: 100_RU
+
+Ukraine|Subgrid 100:
+  definition: Western Ukraine
+  country: Ukrayne
```
danielhuppmann, line 667:
```suggestion
  country: Ukraine
```

nomenclature/definitions/region/subcountries.yaml
```diff
@@ -187,73 +394,310 @@ Italy|Subgrid 98:
...
+
+# ehighway clusters
+
+Russia|Subgrid 100:
+  definition: Russia West of the Urals
+  country: Russia
+  note: This region was defined and used in the ehighway2050 public datasets
+  ehighway2050: 100_RU
+
+Ukraine|Subgrid 100:
```
danielhuppmann, line 665: It seems to me that highways2050 uses a unique identifier for each subgrid - except for Russia and Ukraine, which are both 100. Is that correct?

nomenclature/definitions/region/subcountries.yaml
```diff
@@ -275,33 +719,81 @@ Italy|Subgrid south:
...
-    definition: this 'virtual' region holds the offshore windpower of belgium in norh sea
+    definition: This 'virtual' region represents offshore wind generation of Belgium in the North Sea
     country: Belgium
     notes: This region was defined and used in the plan4res public datasets, it is built as aggregation of ehighway2050 regions
     ehighway2050: 110_NS
     plan4res: NSBE
 
 Germany|North Sea:
-    definition: this 'virtual' region holds the offshore windpower of germany in norh sea
+    definition: This 'virtual' region represents the offshore windpower of germany in the Norh Sea
```
danielhuppmann, line 729:
```suggestion
    definition: This 'virtual' region represents the offshore windpower of Germany in the North Sea
```

sandrinecharousset (approved):

nomenclature/definitions/region/subcountries.yaml
```diff
@@ -275,33 +719,81 @@ Italy|Subgrid south:
...
+    notes: This region was defined and used in the ehighway2050 public datasets
+    ehighway2050: 114_NS
+
+Norway|North Sea|Subgrid 115:
+    definition: This 'virtual' region represents the offshore windpower in the Norh Sea near to Norway
+    country: Norway
+    notes: This region was defined and used in the ehighway2050 public datasets
+    ehighway2050: 115_NS
+
+Sweden|North Sea|Subgrid 116:
```
danielhuppmann, line 791: Wouldn't that be the Baltic Sea?

nomenclature/definitions/region/subcountries.yaml
```diff
@@ -275,33 +719,81 @@ Italy|Subgrid south:
...
+    notes: This region was defined and used in the ehighway2050 public datasets
+    ehighway2050: 109_NS
+
+Denmark|North Sea|Subgrid 113:
+    definition: This 'virtual' region represents the offshore windpower in the Norh Sea near to Denmark
+    country: Denmark
+    notes: This region was defined and used in the ehighway2050 public datasets
+    ehighway2050: 113_NS
+
+Denmark|North Sea|Subgrid 114:
```
danielhuppmann, line 779: Is there a separation of the grid between the North and the Baltic Sea?

danielhuppmann (changes requested):
Sorry, you didn't correct all spelling mistakes - still written "Norh Sea" in a number of region definitions. Also, I noticed some possible confusion between North and Baltic Sea, please double-check.

nomenclature/definitions/region/subcountries.yaml
erikfilias: This region is located at (58°20'04.2"N 10°07'15.2"E). So, it is between Northern Denmark, Southern Norway, and Western Sweden in Map. Thus, I assigned it to Sweden to avoid multiple of these regions assigned to only one country.

danielhuppmann: ok, thanks for double-checking

nomenclature/definitions/region/subcountries.yaml
erikfilias: Yeah, I found both with the same number (100_RU and 100_UA). 
I don't know why about it, really.
It makes no sense for me.

nomenclature/definitions/region/subcountries.yaml
erikfilias: I don't get your question. All these regions are in the North Sea, and no one was defined for the Baltic Sea on eHighway2050. I'm not sure if we'll need to define regions in the Baltic Sea, but it is an open space to discuss.

nomenclature/definitions/region/subcountries.yaml
erikfilias: Notice that both regions have links (according to eHighway2050) that make it difficult to collapse them into only one region. 
For example
- 100_UA:  100_UA - 42_PL (Poland), 100_UA - 58_HU (Hungary), 100_UA - 59_RO (Romania)
- 100_RU:  100_RU - 73_EE (Estonia), 100_RU - 74_FI (Finland), 100_RU - 75_FI (Finland), 100_RU - 77_LT (Lithuania), 100_RU - 78_LV (Latvia)

Moreover, these regions and links are being used in Case Study 3 to build the European power grid's skeleton.

nomenclature/definitions/region/subcountries.yaml
danielhuppmann: If eHighways2050 only looked at locations in the North Sea, no problem - was just wondering...

nomenclature/definitions/region/subcountries.yaml
danielhuppmann: If it is indeed a "bug" or bad definition in eHighways2050, no need for us to worry about it - just wanted to double-check that it isn't a copy-paste error...

erikfilias: Ah, ok, thanks.

danielhuppmann:
There are still a number of typos for region definitions related to the `Norh Sea`

danielhuppmann (approved):
Thanks @erikfilias, now this looks great!

PR #102, opened 2021-02-23
Missing variables of a Pumped-Storage Hydro unit

erikfilias:
This PR adds the following missing variables:

`Initial Storage Level|Electricity|Hydro|Pumped Storage`
`Maximum Charge|Electricity|Hydro|Pumped Storage`
`Maximum Discharge|Electricity|Hydro|Pumped Storage`

And modifies the description of existent pumped-storage hydro's variables.

PR #105, opened 2021-03-02
Additional variables from GENeSYS-MOD for openTEPES and Plan4EU

tburandt:
Additional/changed variables after requests from @erikfilias and @sandrinecharousset

### DIFF
--- a/nomenclature/definitions/variable/energy/energy-final.yaml
+++ b/nomenclature/definitions/variable/energy/energy-final.yaml
@@ -501,6 +501,13 @@ Final Energy|Electricity|Other (excl. Heat, Cooling, Transport):
    sector: all
    energy use: all excl.heating, cooling, transportation               
    unit: GWh/yr
+   
+Final Energy|Electricity|Other (excl. Heat, Transport):
+   description: Final electricity consumption for non heating, non e-transport uses, 
+   energy source: electricity
+   sector: all
+   energy use: all excl.heating, transportation               
+   unit: GWh/yr
 
 Final Energy|Electricity|Profile:
    description: Time series of Final electricity consumption profiles. 
--- a/nomenclature/definitions/variable/technology/power-plant.yaml
+++ b/nomenclature/definitions/variable/technology/power-plant.yaml
@@ -181,19 +181,43 @@ Initial Storage Level|Electricity|Energy Storage System:
    unit: GWh
 
 Maximum Charge|Electricity|Energy Storage System:
-   description: Maximum charge when storing energy the energy storage system unit.
-   unit: MW
+   description: Maximum charge when storing energy in the energy storage system unit.
+   unit: GW
 
 Maximum Discharge|Electricity|Energy Storage System:
    description: Maximum discharge when consuming energy from the energy storage system unit.
-   unit: MW
+   unit: GW
+
+Maximum Charge|Electricity|Energy Storage System|Lithium-Ion:
+   description: Maximum charge when storing energy in the Lithium-Ion Batteries systems.
+   unit: GW
+
+Maximum Discharge|Electricity|Energy Storage System|Lithium-Ion:
+   description: Maximum discharge when consuming energy from the Lithium-Ion Batteries systems.
+   unit: GW
+   
+Maximum Charge|Electricity|Energy Storage System|Redox Flow:
+   description: Maximum charge when storing energy in the Redox Flow Batteries systems.
+   unit: GW
+
+Maximum Discharge|Electricity|Energy Storage System|Redox Flow:
+   description: Maximum discharge when consuming energy from the Redox Flow Batteries systems.
+   unit: GW
+   
+Maximum Charge|Electricity|Energy Storage System|Compressed Air:
+   description: Maximum charge when storing energy in the Compressed air energy storages systems.
+   unit: GW
+
+Maximum Discharge|Electricity|Energy Storage System|Compressed Air:
+   description: Maximum discharge when consuming energy from the Compressed air energy storages systems.
+   unit: GW
 
 Maximum Storage|Electricity|Energy Storage System:
-   description: Maximum energy that can be stored by the energy storage systems.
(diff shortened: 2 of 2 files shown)


nomenclature/definitions/variable/technology/power-plant.yaml
```diff
@@ -181,19 +181,43 @@ Initial Storage Level|Electricity|Energy Storage System:
...
+   
+Maximum Charge|Electricity|Energy Storage System|Compressed Air:
+   description: Maximum charge when storing energy in the Compressed air energy storages systems.
+   unit: GW
+
+Maximum Discharge|Electricity|Energy Storage System|Compressed Air:
+   description: Maximum discharge when consuming energy from the Compressed air energy storages systems.
+   unit: GW
 
 Maximum Storage|Electricity|Energy Storage System:
```
sandrinecharousset, line 215:
```suggestion
Maximum Storage|Electricity|Energy Storage System|Redox Flow:
```

sandrinecharousset (changes requested):

nomenclature/definitions/variable/energy/energy-final.yaml
```diff
@@ -501,6 +501,13 @@ Final Energy|Electricity|Other (excl. Heat, Cooling, Transport):
    sector: all
    energy use: all excl.heating, cooling, transportation               
    unit: GWh/yr
+   
+Final Energy|Electricity|Other (excl. Heat, Transport):
```
danielhuppmann, line 505: The variable name "Other" is usually meant as "everything that doesn't have an own subcategory" - whereas you seem to mean "every category except". Maybe "Total (excl. Heat and Transport)" would be a better identifier?

nomenclature/definitions/variable/technology/power-plant.yaml
```diff
@@ -181,19 +181,43 @@ Initial Storage Level|Electricity|Energy Storage System:
...
-   unit: MW
+   description: Maximum charge when storing energy in the energy storage system unit.
+   unit: GW
 
 Maximum Discharge|Electricity|Energy Storage System:
    description: Maximum discharge when consuming energy from the energy storage system unit.
-   unit: MW
+   unit: GW
+
+Maximum Charge|Electricity|Energy Storage System|Lithium-Ion:
```
danielhuppmann, line 191: It would make sense to start a new type-list like we have for fuels... Also, question is whether pumped-storage should be part of this naming convention or have its own structure, see #102 - @erikfilias, what do you think?

danielhuppmann (changes requested):
Thanks @tburandt - see a few comments inline

nomenclature/definitions/variable/energy/energy-final.yaml
sandrinecharousset: @danielhuppmann @tburandt yes this would indeed make sense

nomenclature/definitions/variable/technology/power-plant.yaml
erikfilias: I also agree with a tag for all types of energy storage. They have in common the following characteristics:
- Maximum Charge
- Maximum Discharge
- Initial Storage
- Maximum Storage
- Minimum Storage
- Efficiency
- Storage type or cycle (daily, weekly, or monthly)
Also, we have the known common variables for all generating unit like:
- Maximum Power
- Minimum Power

So, it can help us to start a new tag. 
Even a hydro unit with a reservoir could be considered for this tag according to an inventory constraint (energy management into an operation cycle or storage type).

erikfilias (approved):

nomenclature/definitions/variable/technology/power-plant.yaml
tburandt: Okay, @danielhuppmann do you think that this should rather be done in a new PR instead of modifying this PR?

sandrinecharousset: @erikfilias from my point of view the variables are different for i/thermal units (the ones for which we have already included the <fuel> tag, ii/energy storage system : yes would be interested to have a <techno> tag (with Lithium-Ion, Compressed Air.....  ii/ Pumped Storage : this is a storage system but it is slightly different, I would not integrate it into the list of energy storage systems at this stage 
Other comment: Maximum power and Minimum Power do not exist for energy storage system, where we have Maximum Charge/Discharge (even if in the end those are equal); efficiency is not there neither (used for hydro, but for energy storage systems there is RoundTrip coefficient)
And finally a Hydro Unit with a reservoir has additionnal characteristics such as inflows... 
We could think of merging all this but it would require discussing it (and would induce some delays in implementation of models as some case studies require that datasets are built quite soon, so we need the nomenclature to be quite stable now, ie what would be great is to limit to minor changes + add variables + correct bugs but not do many major restructurations :-)

erikfilias: I also agree with @sandrinecharousset  that this implementation needs a discussion and induces some delays.
For this, we can continue with this PR and #102.
And to put this new tag as an issue for a while.

danielhuppmann (commented):
Thanks @tburandt for these changes - I tried to resolve the merge conflicts to save you some time, please review and write a comment when it's good to merge.

One observation: @erikfilias changed the units from `GW` (only) to allow both `MW` and `GW` - maybe also relevant in this PR for consistency?

sandrinecharousset (approved):

sandrinecharousset:
> @tburandt @loeffko, this PR has not been active for a year - please review whether the proposed changes are still relevant.
> 
> If yes, please resolve the merge conflicts or start a new branch and PR with the relevant changes. Then, close this PR.

@danielhuppmann @loeffko @tburandt it appears that except maybe for the emissions.yaml file, all other changes are already included in the master or in more recent PRs. I would suggest you review the changes in emissions.yaml and create a new PR if necessary, and close this one.....

PR #115, opened 2021-10-14
Adding <fuel> tag in the variable Emission Rate|CO2|Electricity|*

### LINKED ISSUE #110: harmonize fuel description also in emissions
Just not to forget, I noticed that the file emissions.yaml has not been harminized to the new way of describing the fuels (eg Electricity|<Fuel> at least for the variable Emission rate|Electricity|*
Will do it when I have time, this Issue is here not to forget
https://github.com/openENTRANCE/nomenclature/blob/master/nomenclature/definitions/variable/emissions/emissions.yaml

erikfilias:
This PR modifies the variable Emission Rate|CO2|Electricity|*, including the <fuel> tag and reducing its definition per each fuel.

closes #110


danielhuppmann (approved):
Thanks @erikfilias, great to shorten the code lists!

PR #123, opened 2021-11-05
Add "EU27 & UK"

danielhuppmann:
Adding "EU27 & UK" per request of @Renato-Rodrigues, see  https://github.com/openENTRANCE/nomenclature/issues/119#issuecomment-961889809


danielhuppmann:
Added the new region to the openENTRANCE (public), openENTRANCE (internal) and ECEMF (internal) Scenario Explorer instances.

PR #124, opened 2021-11-08
Adjust unit of prices in economy variables

### LINKED ISSUE #118: Price indices do not have a unit
In the file economy.yaml there are quite some variables that are (price) indices. These price indices have no unit. For example: 

Product Price Index|<Product>:
   description: Price index on <this product> level, relative to base year 2011
   note: The index equals 1 in 2011, the reference year
   unit: ''

Daniel wrote in an email (12-11-2020) the following: 

"I realized when I saw your message that we are having a problem with this kind of units data (where there is no unit) for writing to the database). For now, can you please simply add a „-„ (minus sign) when uploading test scenarios until we fix this?"

However, this minus sign does not work when we try to upload the data. Could we add a unit for all indices where a unit is missing? The name of the unit could simply be 'index'.  

Greets, Paolo and Hettie

hettieboonman:
Prices did not have a unit yet. I have updated the unit of prices 

from: 
unit: ''

to: 
unit: 'index'

closes #118

### DIFF
--- a/nomenclature/definitions/variable/economy/economy.yaml
+++ b/nomenclature/definitions/variable/economy/economy.yaml
@@ -68,17 +68,17 @@ Labor|<Industry>:
 Product Price Index|<Product>:
    description: Price index on <this product> level, relative to base year 2011
    note: The index equals 1 in 2011, the reference year
-   unit: ''
+   unit: 'index'
 
 Capital Price Index:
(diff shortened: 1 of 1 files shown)


danielhuppmann:
Added reference to the corresponding issue in the PR description (so that the issue is closed automatically when merging this PR).

PR #125, opened 2021-11-11
Adding energy charge/discharge variables in power-plants.yaml

erikfilias:
This PR adds the two next variables, related to the energy charge/discharge of the energy storage systems, to the file power-plants.yaml:

Energy Consumption|Electricity|Energy Storage System:
   description: Energy consumption from the power grid to be stored into the ESS.
   unit: [MWh, GWh]

Energy Production|Electricity|Energy Storage System:
   description: Energy discharged from the ESS to be injected into the power grid.
   unit: [MWh, GWh]

CC. @SarahYS


nomenclature/definitions/variable/technology/power-plant.yaml
```diff
@@ -230,4 +230,12 @@ Storage Type|Electricity|Energy Storage System:
 
 Stored Energy Inventory|Electricity|Energy Storage System:
    description: State of the stored energy by ESS
-   unit: [MWh, GWh]
\ No newline at end of file
+   unit: [MWh, GWh]
+   
+Energy Charge|Electricity|Energy Storage System:
```
danielhuppmann, line 235: The "Charge" of a battery (the energy level) is different from "Charging a battery", so this variable name is not ideal. Any suggestions for alternatives? Also, the description should be more specific.

nomenclature/definitions/variable/technology/power-plant.yaml
```diff
@@ -230,4 +230,12 @@ Storage Type|Electricity|Energy Storage System:
...
+   description: Storing energy in the energy storage system unit.
+   unit: [MWh, GWh]
+
+Energy Discharge|Electricity|Energy Storage System:
+   description: Consuming energy from the energy storage system unit.
```
danielhuppmann, line 240: This description is not clear - is it consuming energy from the storage into the grid or the other way around?

danielhuppmann (changes requested):
Thanks @erikfilias - I don't agree with the naming conventions, can you try to think of alternatives?

erikfilias:
> Thanks @erikfilias - I don't agree with the naming conventions, can you try to think of alternatives?

Thank you @danielhuppmann, you were right.
I changed the name of the variable and its description.
Please check it.

sandrinecharousset:
@erikfilias @danielhuppmann shoudn't these variables be in electricity-operation rather than power-plant? ther already exist variables named Active Power|.... (which should be adapted with the <fuel> tag ) there.

erikfilias:
> @erikfilias @danielhuppmann shoudn't these variables be in electricity-operation rather than power-plant? ther already exist variables named Active Power|.... (which should be adapted with the tag ) there.

Many thanks, @sandrinecharousset for the suggestion. I've just made the following:
- Insert the tag <fuel> to the variables in the file "electricity-operation.yaml"
- Moving the proposed variable `Energy Consumption|Electricity|Energy Storage System` from "power-plant.yaml" to "electricity-operation.yaml"
- Add the category "Energy Storage System" to the file "fuel_types.yaml"  

Please, let me know if something is missing or can be improved.

erikfilias:
Hi @danielhuppmann, @sandrinecharousset.
I rebased my local branch and added the changes according to the new version of the yaml files.

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -12,360 +12,82 @@
...
-    description: Operation Cost of  Wind OffShore power plants
-    unit: EUR_2020
+
+- Reserve|Electricity|Automatic Frequency Restoration|<Fuel>:
+    description: Automatic Frequency Restoration Reserve available for <this fuel>
+     power plants; this amount is the maximum available for the grid operator considering
+     the commitment decisions taken by the model
+    unit:
+    - MWh
+    - GWh
```
danielhuppmann, line 79: Is there a particular reason why this variable could not be aggregated by simple summation across regions? ie., why have `skip-region-aggregation: true`

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -12,360 +12,82 @@
...
-    description: Operation Cost of energy storage systems, excl. hydro storage
-    unit: EUR_2020
+
+- Reserve|Electricity|Frequency Containment|<Fuel>:
+    description: Frequency Containment Reserve available for <this fuel>
+     power plants; this amount is the maximum available for the grid operator considering
+     the commitment decisions taken by the model
+    unit:
+    - MWh
+    - GWh
```
danielhuppmann, line 88: As above, why skip region aggregation on this variable?

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -12,360 +12,82 @@
...
-    unit: EUR_2020
-    skip-region-aggregation: true
-- Operation Cost|Electricity|Wind|OnShore:
-    description: Operation Cost of  Wind Onshore power plants
+    unit:
+    - MWh
+    - GWh
+
+- Operation Cost|Electricity|<Fuel>:
+    description: Operation Cost of <this fuel> power plants
```
danielhuppmann, line 69: The tags now work differently because we switched to the stand-alone **nomenclature** package.
```suggestion
    description: Operation Cost of <Fuel> power plants
```

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -12,360 +12,82 @@
...
+    unit:
+    - MWh
+    - GWh
+
 - Active Power|Electricity|Load Curtailment:
     description: Energy curtailed - from the system point of view is equivalent to
       generation of Active Power
-    unit: MWh
+    unit:
+    - MWh
```
danielhuppmann, line 25: Is there a particular reason for allowing multiple units here? Having several units allowed in the database makes working with the data more complicated for users, because they have to keep in mind that orders of magnitude values may differ between different rows.

definitions/variable/tag_fuel_types.yaml
```diff
@@ -74,6 +74,15 @@
   - Coal|Hard coal|w/o CCS:
       description: anthracite and bituminous coal with freely vented CO2 emissions
 
+  - Energy Storage System:
```
danielhuppmann, line 77: Could you please remove this change from this PR and start it as a separate discussion? We should probably distinguish between fuels, energy carriers and power-plant types more carefully.

danielhuppmann (changes requested):
Thank you @erikfilias, see a few suggested changes and questions for clarification

definitions/variable/tag_fuel_types.yaml
erikfilias: Ok, got it.

definitions/variable/technology/electricity-operation.yaml
erikfilias: In case study 3, we are dealing with multiple units. In some way, it is suitable to talk about MWh or GWh instead. I made the changes for coherence.
But I understand your point, and I also agree. 
I'll revert this change.

definitions/variable/technology/electricity-operation.yaml
erikfilias: It could open another discussion; I would prefer not to aggregate according to our model approach. Still, in general terms, it could be used for users to talk about the reserve at different regional aggregations. Then, I'll change these variables to be aggregated.

erikfilias:
@danielhuppmann Thank you for your observations. I tried to attend to every comment. Please check it.

definitions/variable/technology/electricity-operation.yaml
danielhuppmann: Got it, ok

danielhuppmann: You convinced me, you please revert to `skip-region-aggregation: true`

definitions/variable/technology/electricity-operation.yaml
danielhuppmann: As above, please revert to `skip-region-aggregation: true`

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -12,360 +12,71 @@
...
-    description: Active power of Photovoltaic power plants
-    unit: MWh
-- Active Power|Electricity|Wind|OnShore:
-    description: Active power of  Wind Onshore power plants
-    unit: MWh
-- Active Power|Electricity|Wind|OffShore:
-    description: Active power of  Wind OffShore power plants
-    unit: MWh
+- Active Power|Electricity|<Fuel>:
+    description: Active power of <this fuel> power plants
```
danielhuppmann, line 16: Note the slightly changed usage of the tags in the attributes
```suggestion
    description: Active power of <Fuel> power plants
```

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -12,360 +12,71 @@
...
+    unit:
+    - GWh
+
 - Active Power|Electricity|Electricity Storage:
     description: Active Power of energy storage systems, excl. hydro storage
-    unit: MWh
+    unit:
+    - GWh
+
+- Energy Consumption|Electricity|Energy Storage System:
```
danielhuppmann, line 37: Sorry for being a pain here, but I think that this is ambiguous - do you mean the total withdrawal of energy form the grid, and if yes, why is this not just a negative value of *Active Power|Electricity|Electricity Storage*? Or is this the actual consumption for operating the storage facility, ie the losses = difference between withdrawal and injection?

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -12,360 +12,71 @@
...
-      the commitment decisions taken by the model
-    unit: MWh
-- Reserve|Electricity|Frequency Containment|Energy Storage System:
-    description: Frequency Containment Reserve available for energy storage systems,
-      excl. hydro storage; this amount is the maximum available for the grid operator
-      considering the commitment decisions taken by the model
-    unit: MWh
+
+- Reserve|Electricity|Automatic Frequency Restoration|<Fuel>:
+    description: Automatic Frequency Restoration Reserve available for <this fuel>
```
danielhuppmann, line 65: Please note the new usage of the tags in the attributes.
```suggestion
    description: Automatic Frequency Restoration Reserve available for <Fuel>
```

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -12,360 +12,71 @@
...
+     power plants; this amount is the maximum available for the grid operator considering
+     the commitment decisions taken by the model
+    unit:
+    - GWh
+    skip-region-aggregation: false
+
+- Reserve|Electricity|Frequency Containment|<Fuel>:
+    description: Frequency Containment Reserve available for <this fuel>
```
danielhuppmann, line 73: As above
```suggestion
    description: Frequency Containment Reserve available for <Fuel>
```

definitions/variable/technology/power-plant.yaml
```diff
@@ -165,6 +165,11 @@
     description: Efficiency of pumping for a pumped-storage hydro unit
     unit: '%'
     skip-region-aggregation: true
+- Inflows|Electricity|Hydro|Pumped Storage:
+    description: Inflows into a reservoir expressed in energy
+    unit:
+    - MWh
```
danielhuppmann, line 171: Is it necessary to allow both units here? I'd prefer to only allow one, to reduce the risk of errors.

danielhuppmann (commented):
Second round of review comments

definitions/variable/technology/electricity-operation.yaml
erikfilias: Ok, I changed it to `skip-region-aggregation: true`.

definitions/variable/technology/electricity-operation.yaml
erikfilias: Got it.

definitions/variable/technology/electricity-operation.yaml
erikfilias: No worries, I think it is a perfect moment to clarify it.
Commonly, there are two options two to represent the production & consumption of any device that has a storage if I am not wrong:
1.  Two variables to represent the production & consumption separately are also considered in the mathematical formulation of the model.
2. One variable to represent the total withdrawal means `production + losses = consumption`; however, in some cases, the mathematical formulation internally considers production, losses, and consumption as variables.

Then, as we consider the production, losses, and consumption as variables within our models, the natural way is the `option 1` that also as the advantage to show the losses in an explicitly way instead of the `option 2` that the losses are implicit and need to be recovered.

I would prefer option 1, even if it means including an additional variable.
In addition, I'll consult about it to the responsibles in the case study.

definitions/variable/technology/power-plant.yaml
erikfilias: I can keep the `GW` and `GWh` in the respective variables.
I got your point. Thus, I will take off the extra unit and keep only one.

erikfilias:
@danielhuppmann , I tried to take into account all your suggestions.
I think the variables related to the operation of the storage facility are in the table yet. 
I'll wait for your comment about it and try to close it asap.
Thank you!

definitions/variable/technology/electricity-operation.yaml
SarahYS: Not that I am expert on this, but I would vote for option 1 here. (I will also need to use this in our CS)

danielhuppmann: @sandrinecharousset, could you please also comment on this question?

Beyond that, even if we have two variables for injection into the grid and withdrawal from the grid of an electricity storage facility, I think that "Consumption" is a misleading word for it.

definitions/variable/technology/power-plant.yaml
```diff
@@ -194,58 +185,48 @@
...
       systems.
     unit:
-    - MWh
     - GWh
 - Minimum Storage|Electricity|Energy Storage System:
     description: Minimum energy that can be stored by the energy storage system unit.
     unit:
-    - MWh
     - GWh
 - Spillage|Electricity|Energy Storage System:
```
sandrinecharousset, line 224: Shouldn't Spillage be in electricity-operation?

definitions/variable/technology/power-plant.yaml
```diff
@@ -194,58 +185,48 @@
...
     description: Energy spillage that could not be stored by an energy storage system
       (ESS). Several energy models include spillage variables related to renewable
       energy production; however, another way is to include a spillage variable to
       the ESS unit since it is the last sink of the power network.
     unit:
-    - MWh
     - GWh
 - Storage Type|Electricity|Energy Storage System:
```
sandrinecharousset, line 231: If 'type' is related only to seasonality, why not call this variable Seasonality instead of Storage Type?

definitions/variable/technology/power-plant.yaml
```diff
@@ -257,5 +238,4 @@
 - Stored Energy Inventory|Electricity|Energy Storage System:
```
sandrinecharousset, line 238: SHouldn't this be in electricity-operation?

definitions/variable/technology/electricity-operation.yaml
sandrinecharousset: @erikfilias why do you forbid aggregation of reserves? When aggregating 2 regions, the resulting aFR per fuel type is the sum of both, so I wouldn't have prevented this....

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -12,360 +12,71 @@
...
-    unit: MWh
-- Active Power|Electricity|Wind|OffShore:
-    description: Active power of  Wind OffShore power plants
-    unit: MWh
+- Active Power|Electricity|<Fuel>:
+    description: Active power of <Fuel> power plants
+    unit:
+    - GWh
+
 - Active Power|Electricity|Load Curtailment:
```
sandrinecharousset, line 20: We should add also Active Power|Electricity|Load Curtailment|<Demand response tag> together with a new list of categories (we are to do that with CS1 but no one had time yet......)
Same for Load Shifting....

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -12,360 +12,71 @@
...
+    - GWh
+
 - Active Power|Electricity|Load Shifting:
     description: Energy shifted - from the system point of view is equivalent to generation
       of Active Power
-    unit: MWh
+    unit:
+    - GWh
+
 - Active Power|Electricity|Electricity Storage:
```
sandrinecharousset, line 32: shoudn't this be Active Power|Electricity|Electricity Storage System|<type of storage system>
in the scenarios we already have 3 kinds of storage systems with redoxFlow, compressed air.....

definitions/variable/technology/electricity-operation.yaml
sandrinecharousset: But this can be done later......

definitions/variable/technology/electricity-operation.yaml
sandrinecharousset: Argh, I had written a long message but then everything disapeared.......
In summary my feeling is that Active power is enough, but I am open to other views.
For describing an energy storage system (in power-plant.yma), I need Maximum Charge, Maximum Discharge, Maximum Storage, Minimum Storage, Charging efficiency and discharging efficiency, which are the 'physical' characteristcs of the storage;
Then I also need to know the "forced" interactions with the rest of the world, which in my model is summarised in a variable Inflows (a time series), which is positive if energy is being added to the system, negative if it means a demand served by the ESS (not through the grid)
Regarding the operation variables, ActivePower for me is enough if we consider no spillage; But if we consider spillage is possible then we need to have also a Spillage variable. I understand the possibility of having 'consumption' and 'production' but I find this less understandable, and also it would double the size of the data considering that at each timestep you would have both consumption and production instead of only ActivePower...... which in my case (eg in CS1 case where we model appliances with flexibility as storage systems) would mean lots of additionnal rows... (but this is a minor consideration)

definitions/variable/technology/electricity-operation.yaml
erikfilias: > It could open another discussion; I would prefer not to aggregate according to our model approach. Still, in general terms, it could be used for users to talk about the reserve at different regional aggregations. Then, I'll change these variables to be aggregated.

I don't deny that it could be aggregated. We are not using the reserves but I also agree to leave the aggregation as an option. In this sense, I will change to `skip-region-aggregation: False` .

definitions/variable/technology/power-plant.yaml
erikfilias: Got it.

definitions/variable/technology/power-plant.yaml
erikfilias: Got it!

definitions/variable/technology/power-plant.yaml
erikfilias: Got it! I'll move it.

definitions/variable/technology/electricity-operation.yaml
erikfilias: Have you referred to having a new variable like:

- Active Power|Electricity|Load Curtailment|<new list>:

definitions/variable/technology/electricity-operation.yaml
erikfilias: In the beginning, I tried to include the following technologies to `tag_fuel_types.yaml`, in order to be included when the tag <Fuel> is used: 

  - Energy Storage System:
      description: Energy Storage System, excl. hydro storage  
  - Energy Storage System|Lithium-Ion:
      description: Energy Storage System based on the lithium-ion batteries systems
  - Energy Storage System|Redox Flow:
      description: Energy Storage System based on the redox flow batteries systems
  - Energy Storage System|Compressed Air:
      description: Energy Storage System based on the compressed air energy storages systems

@sandrinecharousset, @danielhuppmann, Should we include something similar to `tag_fuel_types.yaml`?

danielhuppmann:
This PR is mixing several items, which makes it hard to review and discuss.

1. Refactoring to use tag-lists - I'll cherry-pick that into a new clean-up PR
2. Changing units from MWh to GWh - is this really necessary?
3. Renaming from "Storage Type" to "Seasonality", but it's not clear how this variable should be used in the first place (what does unit *Daily* mean?)
4. A proposal to add a variable "Energy Consumption", which is seen as not ideal by @sandrinecharousset and myself (because pump-hydro plants to not "consume" electricity)

I suggest to close this PR and start one issue per (relevant) topic for a proper discussion - then start a PR only once an agreement has been found (to avoid having long discussions in PRs, where comments can refer to code that was added and then removed while working - this makes it hard to understand what some comments are about because they are outdated).

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -12,360 +12,84 @@
...
-- Active Power|Electricity|Solar:
-    description: Active power of Photovoltaic power plants
-    unit: MWh
-- Active Power|Electricity|Wind|OnShore:
-    description: Active power of  Wind Onshore power plants
-    unit: MWh
-- Active Power|Electricity|Wind|OffShore:
-    description: Active power of  Wind OffShore power plants
-    unit: MWh
+- Active Power|Electricity|<Fuel>:
```
sandrinecharousset, line 15:
```suggestion
- Active Power|Electricity|{Electricity_Input}:
```
The fuel tag does not include the detail per techonlogy/fuel, only per fuels

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -12,360 +12,84 @@
...
+    description: Active power of <Fuel> power plants
+    unit:
+    - GWh
+
```
sandrinecharousset, line 19:
```suggestion
Active Power|Electricity|Energy Storage System|{Energy Storage System Type}:
   description: Active Power "generated" or "collected" by an energy storage system
   unit: MWh
Active Power|Electricity|Demand Response|Load Shifting|Electricity|Residential|{Residential Load Shifting Asset}:
   description: Active Power "generated" or "collected" within a load shifting process
   unit: MWh
```

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -12,360 +12,84 @@
...
 - Active Power|Electricity|Load Curtailment:
     description: Energy curtailed - from the system point of view is equivalent to
       generation of Active Power
-    unit: MWh
+    unit:
+    - GWh
+
 - Active Power|Electricity|Load Shifting:
```
sandrinecharousset, line 26:
```suggestion
- Active Power|Electricity|Demand Response|Load Shifting:
```

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -12,360 +12,84 @@
...
+    description: Active power of <Fuel> power plants
+    unit:
+    - GWh
+
 - Active Power|Electricity|Load Curtailment:
     description: Energy curtailed - from the system point of view is equivalent to
       generation of Active Power
-    unit: MWh
+    unit:
+    - GWh
```
sandrinecharousset, line 24:
```suggestion
    - [GWh, MWh]
```

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -12,360 +12,84 @@
...
-    unit: MWh
-- Active Power|Electricity|Wind|OffShore:
-    description: Active power of  Wind OffShore power plants
-    unit: MWh
+- Active Power|Electricity|<Fuel>:
+    description: Active power of <Fuel> power plants
+    unit:
+    - GWh
+
 - Active Power|Electricity|Load Curtailment:
```
sandrinecharousset, line 20:
```suggestion
- Active Power|Electricity|Demand Response|Load Curtailment:
```

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -12,360 +12,84 @@
...
-    unit: MWh
+    unit:
+    - GWh
+
 - Active Power|Electricity|Load Shifting:
     description: Energy shifted - from the system point of view is equivalent to generation
       of Active Power
-    unit: MWh
+    unit:
+    - GWh
```
sandrinecharousset, line 30:
```suggestion
    - [GWh, MWh]
```

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -12,360 +12,84 @@
...
+
 - Active Power|Electricity|Electricity Storage:
```
sandrinecharousset, line 32:
```suggestion
- Active Power|Electricity|Energy Storage System:
```

definitions/variable/technology/electricity-operation.yaml
sandrinecharousset: I suggest to remove the Spillage variable and stick to Active power being negative

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -12,360 +12,84 @@
...
+    unit:
+    - GWh
+
+- Spillage|Electricity|Energy Storage System:
+    description: Energy spillage that could not be stored by an energy storage system
+      (ESS). Several energy models include spillage variables related to renewable
+      energy production; however, another way is to include a spillage variable to
+      the ESS unit since it is the last sink of the power network.
+    unit:
+    - GWh
```
sandrinecharousset, line 43:
```suggestion

```

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -12,360 +12,84 @@
...
+
+- Stored Energy Inventory|Electricity|Energy Storage System:
```
sandrinecharousset, line 45:
```suggestion
- Storage|Electricity|Energy Storage System:
```
to be consistant with the other technologies with storage

definitions/variable/technology/electricity-operation.yaml
sandrinecharousset: I suggest to only keep 'Active power' which when positive means the storage system is discharging energy to the grid, and when negative is charging energy from the grid

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -12,360 +12,84 @@
...
+    description: State of the stored energy by ESS
+    unit:
+    - GWh
+
+- Energy Consumption|Electricity|Energy Storage System:
```
sandrinecharousset, line 50:
```suggestion

```

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -12,360 +12,84 @@
...
+    description: Energy consumption from the power grid to be stored into the ESS.
+    unit:
+    - GWh
```
sandrinecharousset, line 53:
```suggestion

```

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -12,360 +12,84 @@
...
     description: Available storage of "Pumped Storage " units
-    unit: MWh
+    unit:
+    - GWh
 - Storage|Electricity|Energy Storage System:
     description: Available storage of energy storage systems, excl. hydro storage
-    unit: MWh
+    unit:
+    - GWh
 - Storage|Electricity|Load Curtailment:
```
sandrinecharousset, line 63:
```suggestion
- Storage|Electricity|Energy Storage System|{Energy Storage System Type}:
    description: Available storage of energy storage systems, excl. hydro storage, per storage type
    unit: GWh
- Storage|Electricity|Load Curtailment:
```

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -12,360 +12,84 @@
...
-    skip-region-aggregation: true
-- Operation Cost|Electricity|Electricity Storage:
-    description: Operation Cost of energy storage systems, excl. hydro storage
+    unit:
+    - GWh
+
+- Operation Cost|Electricity|<Fuel>:
+    description: Operation Cost of <Fuel> power plants
     unit: EUR_2020
     skip-region-aggregation: true
```
sandrinecharousset, line 71:
```suggestion
- Operation Cost|Electricity|{Electricity_Input}:
    description: Operation Cost of <Electricity Input> power plants
    unit: EUR_2020
    skip-region-aggregation: true
 - Operation Cost|Electricity|Energy Storage System|{Energy Storage System Type}:
    description: Operation Cost of {Energy Storage System Type} storages
    unit: EUR_2020
    skip-region-aggregation: true
 - Operation Cost|Electricity|Demand Response|Load Shifting|Residential|{Residential Load Shifting Asset}:
    description: Operation Cost of {Residential Load Shifting Asset} Load shifting assets
    unit: EUR_2020
    skip-region-aggregation: true
- Operation Cost|Electricity|Demand Response|Load Curtailment:
    description: Operation Cost of {Residential Load Shifting Asset} Load shifting assets
    unit: EUR_2020
    skip-region-aggregation: true
```
adding operations costs of storage and demand response

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -12,360 +12,84 @@
...
-      power plants; this amount is the maximum available for the grid operator considering
-      the commitment decisions taken by the model
-    unit: MWh
-- Reserve|Electricity|Frequency Containment|Energy Storage System:
-    description: Frequency Containment Reserve available for energy storage systems,
-      excl. hydro storage; this amount is the maximum available for the grid operator
-      considering the commitment decisions taken by the model
-    unit: MWh
+
+- Reserve|Electricity|Automatic Frequency Restoration|<Fuel>:
```
sandrinecharousset, line 78:
```suggestion
- Reserve|Electricity|Automatic Frequency Restoration|<Electricity_Input>:
```

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -12,360 +12,84 @@
...
+    description: Automatic Frequency Restoration Reserve available for <Fuel>
+     power plants; this amount is the maximum available for the grid operator considering
+     the commitment decisions taken by the model
+    unit:
+    - GWh
+    skip-region-aggregation: false
+
```
sandrinecharousset, line 93:
```suggestion

- Reserve|Electricity|Automatic Frequency Restoration|Energy Storage System|{Energy Storage System Type}:
    description: Automatic Frequency Restoration Reserve available for storage systems
     this amount is the maximum available for the grid operator considering
     the commitment decisions taken by the model
    unit:
    - GWh
    
- Reserve|Electricity|Frequency Containment|Energy Storage System|{Energy Storage System Type}:
    description: Frequency Containment Reserve available for storage systems
     this amount is the maximum available for the grid operator considering
     the commitment decisions taken by the model
    unit:
    - GWh
    
```

sandrinecharousset (changes requested):
@erikfilias

erikfilias:
> @erikfilias

Thank you @sandrinecharousset for all the suggestions.
I rebase the PR's branch, and it's ready to be merged I think.

sandrinecharousset:
> > @erikfilias
> 
> Thank you @sandrinecharousset for all the suggestions. I rebase the PR's branch, and it's ready to be merged I think.

@erikfilias yes I aggree
@danielhuppmann are you ok?

definitions/variable/technology/power-plant.yaml
```diff
@@ -188,69 +176,40 @@
 - Maximum Energy Charge|Electricity|Energy Storage System:
     description: This is the upper bound that energy storage system could charge energy
       at time "t".
-    unit:
-    - MWh
-    - GWh
+    unit: GWh
 - Initial Storage Level|Electricity|Energy Storage System:
     description: Initial energy stored in the energy storage system unit at first
       load level. (ESS that includes hydro, battery, etc.)
```
sandrinecharousset, line 182:
```suggestion
```
I suggest to remove this variable:
1/ this file is meant for techno characteristics of units, which this is not
2/ initial storage with no  date is very ambiguous and in electricity-opeation we already have the variable Storage level (or sth like that) which for the date corresponding to what is 'initial' will have the same content
3/ initial is a case study parameter.....
(i think this was a variable I created so sorry.....)

sandrinecharousset:
> > > @erikfilias
> > 
> > 
> > Thank you @sandrinecharousset for all the suggestions. I rebase the PR's branch, and it's ready to be merged I think.
> 
> @erikfilias yes I aggree @danielhuppmann are you ok?

Sorry I added a last change request (removing a variable which I find useless, ambiguous and redundant...... was created by myself sorry.....)
Once this is merged we need a new PR for including the tag for the energy storage system list (which is still in a PR so I was affraid the tests would fail if I added this already)

PR #127, opened 2021-11-17
Fixing Consumption variable allowed units

Renato-Rodrigues:
Probably a copy paste issue, but instead of `billion v/yr` you should have `billion USD_2010/yr` in the Consumption variable allowed units.


danielhuppmann (approved):
Indeed, thank you for fixing this!

PR #128, opened 2021-11-17
add alternative unit (USD_2010/t CO2) to Price|Carbon variable

Renato-Rodrigues:
add alternative unit (USD_2010/t CO2) to Price|Carbon report to keep consistency with other variables alternatives


danielhuppmann (approved):
Thanks for adding the alternative monetary unit!

Issue #119, opened 2021-11-01
Additional regional aggregations (REMIND)


Renato-Rodrigues:
Request to add support for REMIND regions aggregation for use in the ECEMF project.

Additional regional aggregations:

```
World:
   definition: 'all World countries'

EU27+UK:
   definition: 'EU27 plus United Kingdom' 

UKI:
   definition: 'United Kingdom and Ireland'

NEU:
   definition: 'European countries outside EU (minus UK)'
   
NEN:
   definition: 'Non European Union North countries'
   
NES:
   definition: 'Non European Union South countries'

ECE:	
   definition: 'European Union Center East countries'

ENC:
   definition: 'European Union North Center countries'

ECS:	
   definition: 'European Union Center South'

ESC:	
   definition: 'European Union South Center countries'

ESW:	
   definition: 'European Union South West countries'

EWN:	
   definition: 'European Union West North countries'
```

You can find the detailed mapping of the above mentioned regions in the below file: 
https://github.com/remindmodel/remind/blob/develop/config/21_regions_EU11/regionmapping_21_EU11.csv
region names (column 2), iso-3 codes (column 2):

JohannesEmm:
I agree some region definitions woudl be great to be added esp. also for ECEMF.

Not sure if all models' native modelling regions are required to be added? In case this could be a large super set for all models.

I would also add
-R52 macro regions (region definition see here: https://tntcat.iiasa.ac.at/SspDb/dsd?Action=htmlpage&page=10#regiondefs)


EU27+UK is already implemented as EU28 I think? (and the main analysis dimension to be used in WP1 as I understand?)

sandrinecharousset:
Already existing aggregated regions are listed in aggregate-regions.yaml
Additionnal can of course be added, but please be carefull so that the definition is not ambiguous. We have defined new aggregated regions by listing the already existing defined regions, eg
EU27:
   definition: 'AT + BE + BG + HR + CY + CZ + DK + EE + FI + FR + DE + GR +
   HU + IE + IT + LV + LT + LU + MT + PL + PT + RO + SK + SI + ES + SE + NL'
   notes: 'membership as of February 1, 2020'
Here the list of countries (AT,....) are defined in countries.yaml
Moreover the definition tag is being used in automated python scripts (which with the above example is very easy to do)
As an example from the list above, UKI should be
UKI:
   definition: 'UK+IE'
EU28 could be defined as
EU28:
   definition:'EU27+UK'
Best
[...]

PR #130, opened 2021-11-18
Add region-processing of REMIND 2.1

danielhuppmann:
This PR adds the region-processing to the openentrance workflow, using the REMIND 2.1 model as a prototype. See https://github.com/openENTRANCE/openentrance/issues/119#issue-1041347710 for the related request by @Renato-Rodrigues.

To-do for this PR:
 - [ ] add region-mapping (to be confirmed by @Renato-Rodrigues)
 - [x] add region-processing to workflow
 - [x] add method-attributes to variables in `definitions` that are not simple summation of subregions
   (see #131)


danielhuppmann:
@phackstock, there is a strange error message here as part of the validation:
```
mappings -> REMIND 2.1
31
  'mappings/remind_2.1.yaml' is not in the subpath of '/home/runner/work/openentrance/openentrance' OR one path is relative and the other is absolute. (type=value_error)
```
I tested it locally against the current status of https://github.com/IAMconsortium/nomenclature/pull/51 and it also fails. Can you please take a look?

danielhuppmann:
@Renato-Rodrigues, can you please confirm the following:
 - do you want the renamed native regions of the scenario data to shown as "REMIND 2.1|United Kingdom and Ireland" or "REMIND 2.1|UKI"?
 - that the common-region-aggregation mapping is correct - in particular, should UKI be added to EU27 explicitly?

Renato-Rodrigues:
> @Renato-Rodrigues, can you please confirm the following:
> 
> * do you want the renamed native regions of the scenario data to shown as
>   
>   * `REMIND 2.1|United Kingdom and Ireland`
>   * `REMIND 2.1|UKI`

I would be perfectly fine with either of above options or both. Whatever you find more consistent with the current naming convention on the db.

> * that the common-region-aggregation mapping is correct - in particular, should `UKI` be added to `EU27` explicitly when computing `EU27 & UK`? Not sure whether that would double-count Ireland...

The only change I would make to the `mappings/remind_2.1.yaml` file is setting:

https://github.com/danielhuppmann/nomenclature/blob/abddbf0a0b24663e87ccf88fc04d072bd65c9d49/mappings/remind_2.1.yaml#L29

to
```
- EU27 & UK:
     - DEU
     - ECE
     - ENC
     - ECS
     - ESC
     - ESW
     - EWN
     - FRA
     - UKI
```
Just to explain better, it is more or less the opposite of what you inferred. Currently EU27 in REMIND does not include Ireland, as this country is not disaggregated from UK, that is why we provide an EU27 aggregation in the results we report. During the project course we will split Ireland from UK to be able to provide a perfect EU27 definition directly from the model native regions.

phackstock:
@danielhuppmann ah yes that error occurs when you try to get the relative path of an already relative path. I'll prepare a PR with the fix on nomenclature.

phackstock:
Updated IAMconsortium/nomenclature#51, should be taken care of with that PR.

danielhuppmann:
Thanks for the review @Renato-Rodrigues. I kept the short-names for the renamed native regions for consistency with the REMIND docs.

I implemented the following changes:

0. Change components of "EU27 & UK" as per your comment
1. Remove "EUR" and "NEU" from the native regions saved to the database
2. Remove "NEU" from "R5OECD90+EU"

Open question (related to item 3 above): computing "World" during the scenario processing would be possible, but this would mean that "Price|Carbon" is not included at the global level. This is because pyam currently does not compute weighted averages if the weights are negative - see the discussion at https://github.com/IAMconsortium/pyam/issues/446. So I suggest that for the time being, you compute results at the global level on your side (including the carbon price) and include it in your uploads.

danielhuppmann:
@phackstock, I noticed one design issue of the RegionProcessor - if we do renaming of the native-regions as part of the workflow, we have to apply the region-processing before the validation. On the other hand, doing the region-processing (which is the most time-intensive part of the processing) and only then doing the validation means that we do a lot of computation and only then raise an error if variables or units are incorrect.

Therefore, I split the validation here by dimensions:
1. Variable (and unit) validation
2. Region processing
3. Region validation

Renato-Rodrigues:
> Open question (related to item 3 above): computing "World" during the scenario processing would be possible, but this would mean that "Price|Carbon" is not included at the global level. This is because pyam currently does not compute weighted averages if the weights are negative - see the discussion at [IAMconsortium/pyam#446](https://github.com/IAMconsortium/pyam/issues/446). So I suggest that for the time being, you compute results at the global level on your side (including the carbon price) and include it in your uploads.

@danielhuppmann Then, for now, I will upload World as one of the native REMIND regions. But this bring another request for a change. 
Due to the same reason, I would request to add back `EUR` as one of the REMIND native regions, and remap the `EU27 & UK` region to it, so we can have also Carbon Price results in this regional aggregation level also:
```
- EU27 & UK:
      - EUR
```

I can also do these changes after you merge your pull request. Your choice.

PS: Unfortunately, or fortunately :-), I see the same issue of negative emissions in R5 region aggregations. As Carbon Price is a quite important variable for the global assessments, maybe we will need to focus soon on finding a good compromise about this for the project results.

phackstock:
@danielhuppmann Ah right, good point. I've thought about it for a minute and I think the easiest solution would be to generally limit the `DataStrucutreDefinition.validate()` function to variables (or in the future any other parameters that are not going to be changed in the renaming process). I think that the second validation of the regions is redundant as that is already covered when initializing the `RegionProcessor` under `RegionProcessor.validate_mappings`.

danielhuppmann:
@Renato-Rodrigues, I added `EUR` to the list of native regions it is now directly renamed, so prices are being maintained. But I guess we will need a larger conversation on how to deal with this. The pyam package (which powers this feature) by default does not compute values for weighted average if any weight is negative. This can be overridden with `drop_negative_weights=False` ([read the docs](https://pyam-iamc.readthedocs.io/en/stable/api/iamdataframe.html#pyam.IamDataFrame.aggregate_region)) and it can be passed via the variable-attributes in the definitions - but this can have strange effects, like having a weighted average of negative-infinite prices even if all prices are positive...

danielhuppmann:
@Renato-Rodrigues, I merged as is - please feel free to give this a test-run on the ECEMF-internal Scenario Explorer.

danielhuppmann:
@phackstock, the region-validation as part of `RegionProcessor.apply()` only occurs if a region-mapping exists - models with a region-mapping are not validated there. I'll open an issue in the nomenclature repo, let's continue this discussion over there...

PR #131, opened 2021-11-18
Skip aggregation for variables related to prices, unit-costs, shares, ...

danielhuppmann:
This PR relates to #130, supporting automated region-processing when uploading scenario data to an IIASA Scenario Explorer instance. It adds a "skip-region-aggregation" attribute to all variables that should not be simply summed up over regions (e.g., prices, per-unit costs, shares).


danielhuppmann:
FYI @sandrinecharousset @erikfilias

PR #132, opened 2021-11-19
Add region-mapping for MESSAGEix-GLOBIOM 1.2

danielhuppmann:
This PR adds the region-mapping for the MESSAGEix-GLOBIOM 1.2 based on the mapping used in AR6.

@behnam-zakeri, in addition to the R5 regions, I also added `EU27 & UK` composed of EEU and WEU - or do you think that mapping this to only WEU makes more sense?


behnam-zakeri:
Thanks, @danielhuppmann, for proceeding with this. The aggregate of WEU and EEU from MESSAGEix-GLOBIOM is more consistent with `Europe (incl. Turkey)` if that is possible. Because EEU includes non-EU countries in West Balkan, and WEU includes Turkey.

danielhuppmann:
Thanks @behnam-zakeri, changed as suggested. Guess we will need some method so that users can easily see which (global) models can provide certain levels of European aggregates - but not for this PR...

PR #133, opened 2021-11-29
Adding support for MEESA v.1.1 native model regions

Renato-Rodrigues:
This pull request needs to be revised be the MEESA team.


danielhuppmann:
Thanks @Renato-Rodrigues for starting the PR! The error messages from the GitHub Actions workflow are a bit difficult to read at first, so the relevant lines are:

```
Error parsing file mapping values are not allowed here
  in "mappings/meesa_v1.1.yaml", line 3, column 8
Error parsing file while parsing a block collection
  in "definitions/region/model_native_regions/meesa_v1.1.yaml", line 1, column 2
expected <block end>, but found '<block sequence start>'
  in "definitions/region/model_native_regions/meesa_v1.1.yaml", line 2, column 3
```

Basically, the yaml files cannot be parsed correctly...

Also, there seems to be some confusion in the mappings-file, the entry "common_regions" exists twice...

Renato-Rodrigues:
The second error is fixed. I am not sure what causes the first one although.

Renato-Rodrigues:
@danielhuppmann  It should be fixed now. There was a rogue space in the yaml file.

tatar99 (approved):
I think it is ok now.
IT

danielhuppmann:
Thanks for the combined effort, @Renato-Rodrigues & @tatar99!

PR #134, opened 2021-12-01
adding alternative euro 2020 units to technology variables

Renato-Rodrigues:


Renato-Rodrigues:
I followed the example of other variables and used millions when defining euro units instead of billions, as it is set for dollar units. I am not 100% sure of that although.

danielhuppmann (approved):
Thanks @Renato-Rodrigues - agree that using million makes for sense as an order of magnitude...

PR #136, opened 2021-12-02
Add additional unit in energy-final.yaml

sebastianzwickl:
This PR includes the following minor edits in the _energy-final.yaml_ file.

- Adding the unit `GWh` 
- Replacing `EJ/yr` by `EJ` (temporal information is already in the time, year, or sub-annual column in the common format)

### DIFF
--- a/definitions/variable/energy/energy-final.yaml
+++ b/definitions/variable/energy/energy-final.yaml
@@ -15,34 +15,44 @@
     energy source: all
     energy use: all excl. losses
     sector: all end-users
-    unit: EJ/yr
+    unit: 
+    - EJ
+    - GWh
 - Final Energy|Electricity:
     description: Final energy consumption of electricity (including on-site solar pv),
       excluding transmission/distribution losses
     energy source: electricity
     energy use: all excl. losses
     sector: all end-users
-    unit: EJ/yr
+    unit: 
+    - EJ
+    - GWh
 - Final Energy|Fossil:
     description: Final energy consumption from fossil fuels
     energy source: fossil fuels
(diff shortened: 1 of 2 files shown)


definitions/variable/energy/energy-final.yaml
```diff
@@ -15,416 +15,540 @@
     energy source: all
     energy use: all excl. losses
     sector: all end-users
-    unit: EJ/yr
+    unit: 
+    - EJ
+    - GWh
```
danielhuppmann, line 20: Same for all suggested changes
```suggestion
    unit: [EJ/yr, GWh]
```

danielhuppmann (commented):
Looking at this PR, I'm a bit worried that this will break consistency with ongoing efforts in AR6, NAVIGATE, ENGAGE, etc. where teams use "EJ/yr".

My (slightly awkward) suggestion: use "GWh" for subannual resolution and "EJ/yr" for yearly data, to clarify that in a five-yearly or decadal timeseries, teh value describes one year (e.g. 2020) and not the time period until the next data point (not the cumulative value for 2020-2029).

Note that this requires a two-step-conversion with pyam, but is still reasonable, I guess:
```python
df.rename(unit:{"GWh": "GWh/yr").convert_unit( "GWh/yr", "EJ/yr")
```

@sebastianzwickl, if you agree, can you add a guidance note at the top of this document?

definitions/variable/energy/energy-final.yaml
sebastianzwickl: Thank you Daniel, I will add this suggested change.

PR #137, opened 2021-12-03
Adding support for ECEMF priority 1 variables

Renato-Rodrigues:
This pull request adds support for missing variables to comply with the ECEMF WP1 priority 1 variables list.

### DIFF
--- a/definitions/variable/energy/energy-final.yaml
+++ b/definitions/variable/energy/energy-final.yaml
@@ -384,17 +384,17 @@
 - Final Energy|Transportation|Fossil:
     description: Final energy consumption by the transportation sector of fossil fuels
     energy source: fossil
-    energy use: freight transportation
+    energy use: transportation
     sector: all
     unit: [EJ/yr, GWh]
 - Final Energy|Transportation|Passenger:
-    description: Final energy consumed for passenger transportation 
+    description: Final energy consumed for passenger transportation, excluding bunkers 
     energy source: all
-    energy use: freight transportation
+    energy use: passenger transportation
     sector: all
     unit: [EJ/yr, GWh]
 - Final Energy|Transportation|Freight:
-    description: Final energy consumed for freight transportation
+    description: Final energy consumed for freight transportation, excluding bunkers 
     energy source: all
     energy use: freight transportation
     sector: all
@@ -540,3 +540,16 @@
     energy use: transportation
     sector: all
     unit: [EJ/yr, GWh]
+ - Final Energy|Bunkers|Passenger:
+    description: Final energy consumed for passenger transportation in international bunkers 
+    energy source: all
+    energy use: passenger transportation
+    sector: all
+    unit: [EJ/yr, GWh]
+- Final Energy|Bunkers|Freight:
+    description: Final energy consumed for freight transportation in international bunkers 
+    energy source: all
+    energy use: freight transportation
+    sector: all
+    unit: [EJ/yr, GWh]
+    
--- a/definitions/variable/energy/energy-final.yaml
+++ b/definitions/variable/energy/energy-final.yaml
@@ -189,15 +189,39 @@
     energy use: none
     sector: all
     unit: EJ/yr
-- Final Energy|Non-Energy Use|Gas:
-    description: Final energy consumption of gas by the non-combustion processes
+- Final Energy|Non-Energy Use|Gases:
+    description: Final energy consumption of gases by the non-combustion processes
     energy source: gases
+    energy use: none
+    sector: all
+    unit: EJ/yr
+- Final Energy|Non-Energy Use|Hydrogen:
+    description: Final energy consumption of hydrogen by the non-combustion processes
+    energy source: hydrogen
+    energy use: none
+    sector: all
+    unit: EJ/yr
+- Final Energy|Non-Energy Use|Liquids:
+    description: Final energy consumption of liquids by the non-combustion processes
+    energy source: liquids
     energy use: none
     sector: all
     unit: EJ/yr
 - Final Energy|Non-Energy Use|Oil:
     description: Final energy consumption of oil by the non-combustion processes
     energy source: oil
+    energy use: none
+    sector: all
+    unit: EJ/yr
+- Final Energy|Non-Energy Use|Solids:
+    description: Final energy consumption of solids by the non-combustion processes
+    energy source: solids
+    energy use: none
+    sector: all
+    unit: EJ/yr
+- Final Energy|Non-Energy Use|Other:
+    description: Final energy consumption of other feedstocks by the non-combustion processes
+    energy source: other
     energy use: none
     sector: all
     unit: EJ/yr
@@ -360,6 +384,12 @@
     energy use: freight transportation
     sector: all
     unit: EJ/yr
+- Final Energy|Transportation|Passenger:
+    description: Final energy consumed for passenger transportation 
+    energy source: all
+    energy use: freight transportation
+    sector: all
+    unit: EJ/yr
 - Final Energy|Transportation|Freight:
     description: Final energy consumed for freight transportation
     energy source: all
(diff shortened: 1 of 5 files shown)


definitions/variable/energy/energy-final.yaml
```diff
@@ -467,3 +497,43 @@
       energy.
     sector: all
     unit:
+
+- Final Energy|Other:
+    description: Final energy consumption of other energy sources that do not fit to any
+      other category
+    unit: EJ/yr
+
+- Final Energy|Transportation|Bunkers:
```
danielhuppmann, line 506: In a parallel discussion with @robertpietzcker and others in the NAVIGATE project, we agreed to move "Bunkers" to the same level as "Transportation".

definitions/variable/energy/energy-secondary.yaml
```diff
@@ -1,11 +1,18 @@
 # List of variables related to the secondary energy sectors
 
+- Secondary Energy:
+    description: Total secondary energy
```
danielhuppmann, line 4: Please provide a clear description what is counted here. If I turn crude oil into diesel and then burn it in a powerplant, is this counted twice?

danielhuppmann (commented):
I suggest to use this opportunity to refactor the final-energy-sector to use the tag system (see the file `tag_transport_types.yaml` as a template). Note that each tag has to be in its own file starting with `tag_`.

Tag `Energy Carrier`:
 - Fossil
 - Electricity
 - Gases
 - Geothermal
 - Heat
 - Hydrogen
 - Liquids
 - Liquids|Biomass
 - Liquids|Other
 - Solids
 - Solids|Biomass
 - Solids|Biomass|Traditional
 - Solids|Biomass|Modern
 - Solids|Coal
 - Other

Tag `Sector`:
- Industry
- Residential and Commercial
- Transportation

"Bunkers" and "Non-Energy Use" are also sectors here, but have a different sub-categorization...

This makes extending the list of energy-carriers and sectors much easier, because **nomenclature** will automatically generate all combinations. It also helps to keep the various sub-trees between different top-level categories (e.g., `Final Energy` and `Price|Final Energy` consistent.

Then, the definition of the price variables can be simplified to
```
- Price|Final Energy|<Sector>|<Energy Carrier>:
    description: End-user price for <Energy Carrier> in the <Sector> sector. 
    notes: Prices should include the effect of carbon prices and all other taxes.
    unit: [EUR_2020/GJ, USD_2010/GJ, EUR_2020/kWh]
    skip-region-aggregation: true
```

This could then also be implemented in the file `energy-final.yaml`.

Renato-Rodrigues:
> I suggest to use this opportunity to refactor the final-energy-sector to use the tag system (see the file `tag_transport_types.yaml` as a template). Note that each tag has to be in its own file starting with `tag_`.

I agree, but due to the time pressing issue for the model partners results upload before holidays, I would ask to first accept the current pull request, and then start working on the more structured alternative.

definitions/variable/energy/energy-secondary.yaml
Renato-Rodrigues: Maybe I am not the best person to answer that, but I would assume that to be consistent you need to report only the net secondary energy use here so, for example, if you produce a secondary energy to be used in another secondary energy process, this should be discounted from the total sum. 
Isn't there already an standard for that variable from other projects and/or reportings?

definitions/variable/energy/energy-final.yaml
Renato-Rodrigues: In Final energy this is clear, however I would like to clarify how to handle this in the Emissions variables.

I would suggest reporting bunkers in the root of the emissions categories (option 1. Emissions|CO2|Bunkers), instead of the same level as transportation (option 2. Emissions|CO2|Energy|Demand|Bunkers).

In that way, all energy aggregations are by default provided without bunkers, which is much more useful in national results, and bunkers can be sum up or not if you want to include them in the analysis. 
Option 1 would also allow direct comparison with UNFCCC/IPCC tables as far as I know.  

Did you discussed that also, or reached a decision about it?

Renato-Rodrigues:
> Tag `Energy Carrier`:
> 
> * Fossil
> * Electricity
> * Gases
> * Geothermal
> * Heat
> * Hydrogen
> * Liquids
> * Liquids|Biomass
> * Liquids|Other
> * Solids
> * Solids|Biomass
> * Solids|Biomass|Traditional
> * Solids|Biomass|Modern
> * Solids|Coal
> * Other

On the energy carrier, would you agree with this list? Filling the gaps from the above list you suggested:

image

definitions/variable/energy/energy-secondary.yaml
danielhuppmann: In the ENGAGE project, I found the following:

> total secondary energy - the sum of all secondary energy carrier production (for consistency checks)

Not sure if this is particularly helpful...

Renato-Rodrigues: I am fine with that alternative, as this is what we do currently in REMIND.

definitions/variable/energy/energy-final.yaml
danielhuppmann: If I understand the notes from the meeting with @robertpietzcker and @gunnar-pik correctly, their proposal was to include CO2 emissions from bunkers as `Emissions|CO2|Energy|Demand|Bunkers`, not as you suggest like `Emissions|CO2|Bunkers`... Any comments?

danielhuppmann:
Thanks for completing the list.

- I'd keep "Fossil" for easy comparison with previous studies.
- I'd keep "Other" and not rename to "Other carrier"
- What was your rationale for adding "Solar", and how do you think it should be defined?
- Not sure what you mean with "Liquids|Electricity"?
- Not sure if adding the types of fossil fuels is necessary at this point.

Please note that #136 added GWh as a unit in this file for all variables, which is now creating a merge conflict. Could you please rebase this PR on the latest `main` branch?

Renato-Rodrigues:
> * I'd keep "Fossil" for easy comparison with previous studies.

sure. I don't see many models having only aggregated values for fossils, due to the differences on emission factors for example, but it does not hurt to keep this category for consistency with previous uploaded data. 

> * I'd keep "Other" and not rename to "Other carrier"

sure. I think "other carrier" is more descriptive for the variable, as we already have "other" all over the place in many variables, but I leave at your discretion the decision, as keeping backwards compatibility is also desirable.

> * What was your rationale for adding "Solar", and how do you think it should be defined?

I believe this category is necessary, and it is currently included also in the variables template, due to final energy consumption of solar energy (e.g., from roof-top solar hot water collector systems). 

> * Not sure what you mean with "Liquids|Electricity"?

These correspond to final energy consumption of synthetic liquid efuels ("power-to-liquid")

> * Not sure if adding the types of fossil fuels is necessary at this point.

Liquids|Fossils, Gases|Fossils and Solids|Fossils are necessary to represent fuel from conventional fossils, against fuels that come from synthetic fuels (Liquids|Electricity, Gases|Electricity and Solids|Electricity).

Also, most models do not have at the final energy level carriers sub-split by primary energy source (ex: Liquids|Coal, Liquids|Gas, Liquids|Oil), as these transformations are usually carried at the secondary energy level, and accounting for them at final energy level require a more complex calculation that take into consideration the transformation and losses at secondary energy level.
Therefore the Fossils subsplit would be more general than the alternative ones that have the link to primary energy (the ones that I underlined under blue in the table).   
 
> Please note that #136 added GWh as a unit in this file for all variables, which is now creating a merge conflict. Could you please rebase this PR on the latest `main` branch?

Sure, I will adapt to that change, although I usually think that GWh make more sense as alternative units only on electricity variables, and not generally.

definitions/variable/energy/energy-final.yaml
gunnar-pik: Agreed - it should be  `Emissions|CO2|Energy|Demand|Bunkers`, so that it's in the same level as transportation

definitions/variable/energy/energy-secondary.yaml
robertpietzcker: My understanding until now is that yes, total SE is the sum of all SE productions - so if you convert electricity to H2 to el to H2, you continually increase the SE amounts. 
This is one of the reasons why SE values can be (much) larger than FE values. 
I think if we would switch to a net accounting of SE, we would loose tractability and information - we already have the FE reporting to take note of what finally arrives at an end user.
But yes, I agree, the original PE/SE/FE split was much nicer before there were SE2SE conversions

robertpietzcker:
> > * I'd keep "Other" and not rename to "Other carrier"
> 
> sure. I think "other carrier" is more descriptive for the variable, as we already have "other" all over the place in many variables, but I leave at your discretion the decision, as keeping backwards compatibility is also desirable.

I would side with "other carrier" here - "other" is confusing as it leaves open whether it refers to sector or carrier. This is clearly true for people looking at the variables, but as many examples from past MIPs show, this is even confusing to the modelers who sometimes mis-assign their model output

definitions/variable/energy/energy-final.yaml
Renato-Rodrigues: Thks. I changed a variable and a description in [this](https://github.com/openENTRANCE/openentrance/pull/137/commits/6c77a37c49d59ec07d30f866916ca7e70df0ca1f) commit to be compatible with this bunker emissions structure.

The final energy bunker variables are also compatible with what was discussed here ([commit](https://github.com/openENTRANCE/openentrance/pull/137/commits/6869bb47e6f0b95788bc57e574e7da846f7ba4df)).

I believe the pull request in its current state is compatible with all comments made so far here, please correct me if I am wrong (except for the tag system use that still needs to be implemented).
@danielhuppmann Would you agree on merging it so we can have a working framework for WP1 ECEMF results submission? Meanwhile I will work in parallel to submit a new pull request that makes use of the tag system whenever possible.

danielhuppmann (approved):
Thanks @Renato-Rodrigues and @robertpietzcker - merging now, and will then refactor to use tag-lists instead of explicit repetition.

PR #140, opened 2021-12-13
Refactor to use `nomenclature.process`

danielhuppmann:
This PR refactors the workflow to use the new `nomenclature.process`.

FYI @phackstock.


phackstock:
Very nice, always a good sign when a PR has more deletions than additions while keeping the same functionality.

PR #142, opened 2021-12-21
added <fuel> support for SE|Hydrogen, biomass/coal split for SE|Solids

robertpietzcker:


danielhuppmann (approved):
Looks good, thanks @robertpietzcker

robertpietzcker:
removed <fuel> implementation again, instead added the explicit gas/coal versions.
(was unsure about implications of the <fuel> thing, need more time to understand that

robertpietzcker:
oh, I thought I could do a second (separate) pull request with the unit changes for the economy. sorry that this got automatically included in the first pull request.

definitions/variable/economy/economy.yaml
```diff
@@ -2,69 +2,69 @@
 
 - Consumption:
     description: Total consumption of all goods, by all consumers in a region
-    unit: [million EUR_2020/yr, billion USD_2010/yr]
+    unit: [million EUR_2020/yr, billion USD_2010/yr, billion EUR_2020/yr]
```
danielhuppmann, line 5: Let us please use only either million EUR or billion EUR - having the same unit at different orders of magnitude is basically an invitation for confusion...

definitions/variable/energy/energy-secondary.yaml
```diff
@@ -56,6 +56,24 @@
 - Secondary Energy|Hydrogen|Fossil|w/o CCS:
     description: Hydrogen production from fossil fuels with freely vented CO2 emissions
     unit: EJ/yr
+- Secondary Energy|Hydrogen|Coal:
```
danielhuppmann, line 59: Might as well use "Secondary Energy|Hydrogen|<Fuel>" and remove the per-fuel variables...

danielhuppmann (changes requested):
Two comments inline.

definitions/variable/economy/economy.yaml
robertpietzcker: totally fine with this - I was afraid that if I remove something, I would interfere with the use of the nomenclature in another project. 
I take your comment as implying that I can remove the million EUR unit, so I will do a commit with this

definitions/variable/energy/energy-secondary.yaml
robertpietzcker: I have to admit I don't fully understand the comment?
or do you mean I should simply commit "Secondary Energy|Hydrogen|<Fuel>" and remove the explicit types?

danielhuppmann: yes, please re-insert `Secondary Energy|Hydrogen|<Fuel>` and delete all explicit hydrogen-fuel variables (double-checking that the fuel is included in the Fuel list).

danielhuppmann (approved):
Thanks @robertpietzcker!

PR #143, opened 2021-12-27
Updating REMIND native regions

Renato-Rodrigues:
* adding EUR and World REMIND native region aggregations
* fixing REMIND regions mapping


mappings/remind_2.1.yaml
```diff
@@ -22,8 +22,9 @@ native_regions:
   - SSA: REMIND 2.1|SSA
   - UKI: REMIND 2.1|UKI
   - USA: REMIND 2.1|USA
-  - EUR: EU27 & UK
-  - World
+  - EU27 & UK: REMIND 2.1|EUR
+  - NEU: REMIND 2.1|NEU
+  - World: REMIND 2.1|World
```
danielhuppmann, line 27: The current implementation saves the "World" region as a common-region. The proposed change would save it (only) as a REMIND-native region, so it would not be possible to easily compare the global aggregates between models.

In principle, the model-specific identifier "<Model>|..." should only be used where we see a risk of confusion between different models. The definition of "World" should be pretty much identical across models, so no need to add the model-identifier.

```suggestion
```

definitions/region/model_native_regions/remind_2.1.yaml
```diff
@@ -21,3 +21,6 @@
   - REMIND 2.1|SSA
   - REMIND 2.1|UKI
   - REMIND 2.1|USA
+  - REMIND 2.1|EUR
+  - REMIND 2.1|NEU
+  - REMIND 2.1|World
```
danielhuppmann, line 26: See comment below.
```suggestion
```

danielhuppmann (changes requested):
Thanks - see comments inline...

mappings/remind_2.1.yaml
```diff
@@ -22,8 +22,9 @@ native_regions:
   - SSA: REMIND 2.1|SSA
   - UKI: REMIND 2.1|UKI
   - USA: REMIND 2.1|USA
-  - EUR: EU27 & UK
-  - World
+  - EU27 & UK: REMIND 2.1|EUR
```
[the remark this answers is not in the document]
Renato-Rodrigues: I must confess I did not understood 100% the mapping files. Could you point me to a documentation or explain me what is the meaning of what is before and after the colon in both files: mappings/remind_2.1.yaml and definitions/region/model_native_regions/remind_2.1.yaml 

 - `EUR` is the native region in REMIND and this should be renamed to `EU27 & UK` in the iiasa database.
 - I am not using automated aggregates for the EUR region because the need to include carbon price and other prices in the results, and as far as I understood the iiasa database does not support weights for the prices aggregations, am I wrong?

mappings/remind_2.1.yaml
Renato-Rodrigues: Following the same principle as "World", shouldn't we need to modify DEU and FRA mappings to have their values available on the model comparisons? They are single country results that are available directly from REMIND native regions.

mappings/remind_2.1.yaml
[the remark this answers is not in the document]
danielhuppmann: See the documentation of the model-region-mapping files [here](https://nomenclature-iamc.readthedocs.io/en/latest/usage.html#regionprocessor).

pyam (and therefore the database processing infrastructure) does support weighted average, but there was no decision about prices should be handled in the ECEMF project (see my email on Nov 19 and following emails).

mappings/remind_2.1.yaml
danielhuppmann: Well, I would treat all regions other than "World" like this:

 - include every model-native region as "<Model>|<Native Region>" in the database, i.e., if a model has 28 regions, there should be 28 entries in the region-definitions files and (at least) 28 entries in the native-region mapping file.
 - if a native-region and a common-region are identical, duplicate it by adding a common-region mapping, in your case
    ```yaml
   - Germany:
        - DEU
    ```
 - for uploading scenario data directly at a common-region level (for example because automated region-processing does not support a particular computation), the native-region-renaming feature can be used, i.e,, 
    ```yaml
   - EUR: EU27 & UK
    ```

Following this logic, a user (in the Scenario Explorer or working with an exported scenario ensemble data file) can immediately find all native regions from every model by taking all regions that start with the model-identifier.

There is a trade-off here between duplication of data and user-friendliness, and I would go for user-friendliness (except for World, where it is obvious that all global models report these values and there  are very little ambiguities about the definition).

danielhuppmann:
Thanks, looks good to me! Just for clarity, though, initially in this PR, you added a new native region "NEU", but in the latest commit, you removed it again - was that on purpose?

Also, could you add a comment (a line starting with `#`) at the top of the remind_2.1.yaml file pointing to the model documentation, in particular the regions, so that interested users can easily find a translation from "ECS", etc. to a human-readable region name?

Renato-Rodrigues:
> Thanks, looks good to me! Just for clarity, though, initially in this PR, you added a new native region "NEU", but in the latest commit, you removed it again - was that on purpose?

NEU is an internal region aggregation made by the REMIND team to allow comparisons between REMIND with 12 regions and REMIND with 21 regions.
I removed it from the mapping as it seems of not much use from ECEMF perspective.

danielhuppmann:
Thank you for the clarification!

danielhuppmann (approved):
Thanks @Renato-Rodrigues

PR #144, opened 2022-01-06
Euro calliope native regions and mapping

FLomb:
Following the indications provided in "ECEMF T1.1 Model Diagnostic Exercise – Protocol v2.0", I have added Euro-Calliope 2.0's native regions and their mapping to larger aggregate groups of regions (e.g. EU27). The names I am using for regions in the YAML files in this pull request are the same I am using for the upload in the ECEMF's Scenario Explorer.

### DIFF
--- a/definitions/region/model_native_regions/euro-calliope_2.0.yaml
+++ b/definitions/region/model_native_regions/euro-calliope_2.0.yaml
@@ -1,3 +1,5 @@
+# The model documentation for Euro-Calliope 2.0 can be found at https://sentinel.energy/wp-content/uploads/2021/03/D4.2-EC.pdf (pp. 9-32)
+# Further details are provided at https://papers.ssrn.com/sol3/papers.cfm?abstract_id=4012180 and https://github.com/calliope-project/euro-calliope-2.0
 - Euro-Calliope 2.0:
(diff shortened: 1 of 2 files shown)


definitions/region/model_native_regions/euro-calliope_2.0.yaml
```diff
@@ -0,0 +1,36 @@
+- Euro-Calliope 2.0:
```
danielhuppmann, line 1: Please add a comment with a reference to the model documentation, in particular the region list.
See this example: https://github.com/openENTRANCE/openentrance/blob/main/definitions/region/model_native_regions/remind_2.1.yaml

danielhuppmann:
Please take a look at my inline review comment to add some reference to the region documentation of the model.

Also, nomenclature-validation is currently failing because "EU27 (excl. Malta)" and "EU27+UK (excl.Malta)" are not defined regions for this project (see definitions/regions/european-regions.yaml for available aggregate-regions).

FLomb:
Thanks @danielhuppmann, all is clear. I'll be back to the PR asap

danielhuppmann (approved):
Thanks @FLomb

PR #147, opened 2022-01-20
EU27 & EFTA

HauHe:
This pull requests adds the region `EU27 & EFTA` to the list of European regions.
Currently this region covers all countries in the EU ETS.


definitions/region/european-regions.yaml
```diff
@@ -27,6 +27,10 @@
       definition: IS + LI + NO + CH
       notes: European Free Trade Association
       source: https://en.wikipedia.org/wiki/European_Free_Trade_Association
+  - EU27 & EFTA:
+      definition: EU27 + EFTA
+      notes: In 2021 EU27 + EFTA comprise all countries that are participating in the EU ETS.
```
danielhuppmann, line 32: Suggestion for an edit to highlight that the UK is not part of this region (or EU ETS)...
```suggestion
      notes: This region is the jurisdiction of the EU ETS as of 2021 (post-Brexit).
```

danielhuppmann (approved):
Thanks @HauHe!

PR #148, opened 2022-01-20
Addin native regions and regions mappings for OSeMBE

HauHe:


definitions/region/model_native_regions/osembe_v1.0.0.yaml
```diff
@@ -0,0 +1,31 @@
+- OSeMBE v1.0.0:
```
danielhuppmann, line 1: change the hierarchy name and all region names as well...
```suggestion
- OSeMBE v1.0:
```

danielhuppmann (commented):
Thanks for the PR, @HauHe! One suggestion, though, to make life easier going forward - I guess the region-mappings will not change between versions 1.0.0 and 1.0.1 of the model. Instead, if the regional dimension of a model changes, this should be (at least v1.1) or even better v2.0.0.

So to avoid that we have to frequently add new-but-identical regions, I suggest to change the files names and the region names to `v1.0` (remove trailing `.0`) or even `v1` if you think you a region change is unlikely before you release v2.

Only leave pointing to the exact patch-version of the model
```
model: OSeMBE v1.0.0
```
and we can extend the **nomenclature-iamc** package such that the `model` attribute can accept a list.
This way, adding a new minor/patch-version of the model, we only have to add the new version to the model-attribute in the model-mapping file, and not add almost-identical regions. This should make it more user-friendly working with the data, too.

cc @phackstock

HauHe:
Hi @danielhuppmann ,
thanks for the comment and suggestion! I will change OSeMBE v1.0.0 to OSeMBE v1.0, then it also in line with the other models in the comparison.

HauHe:
One more thing, if we use OSeMBE v1.0 here instead of OSeMBE 1.0.0 does this also need to be changed in the database of the scenario explorer @danielhuppmann ?

danielhuppmann:
Yes, but the regions are only added to the database infrastructure after this PR is merged, so no harm yet - so that's why we're having these discussions now...

HauHe:
I think now the PR is ready to go. Or am I missing somethin @danielhuppmann ?

danielhuppmann (approved):
Thanks @HauHe!

PR #151, opened 2022-02-02
EU27 & EFTA added to OSeMBE native regions

HauHe:
This PR adds 'EU27 & EFTA' as a native region to OSeMBE. This will allow to report results that are diretly for this region and don't require aggregation.
I'm not sure if this will create any issues with the region mapping and the calculations when summing variables for regions in a mapped region.


danielhuppmann:
Thanks for the PR, but not sure if this really makes sense...

To clarify how this would work: if merged, uploading a scenario will result in two entries for each scenario:
 - a region `OSeMBE v1.0|EU27 & EFTA`, which is a direct copy of what is uploaded as region `EU27 & EFTA`
 - a region `EU27 & EFTA` which is aggregated in the infrastructure

I fear that this could be confusing to users, because it will not be obvious to them that there are two regions with almost-similar names, and they will not understand what the difference is.

A better solution would be to adapt the nomenclature-package such that it allows direct submissions (ie upload as `EU27 & EFTA`) and computes the aggregation automatically only for those variables that are not included directly, and stores all this data under the same region name. But this would require some work and realistically, we do not have resources to implement this before March.

A quick-fix would be to not do the automated aggregation at all, and submit all data for `EU27 & EFTA` as a "native region" (without renaming it to a model-specific region).

HauHe:
Hi Daniel,
I should have probably explained a bit better why I did this. 
In OSeMBE we work in some scenarios with an emission limit. It covers usually the EU27 & EFTA countries since they are part of the EU ETS. 
For these scenarios we would like to report the dual value of the emission limit as the price for emission. And since the emission limit does apply to the region as one I needed 'EU27 & EFTA' in the native regions.

I didn't realise that the solution I proposed would imply two different region names. But I agree that that would create confusion. 
I'm not sure which of your proposed solutions is better. I guess for the model comparisson the agggregates might be quite useful. And even though the carbon price is an important variable it might be less important than the aggregates? Not sure. What do you think?

danielhuppmann:
Yes, the use case absolutely makes sense, we just need to figure how best to support it. In the meantime, please take a look at https://nomenclature-iamc.readthedocs.io/ and let us know (or start a PR) for improving the docs!

I started a PR in the nomenclature-package repo, see https://github.com/IAMconsortium/nomenclature/issues/96. Pinging @phackstock, let's discuss if/how/when we can implement this...

danielhuppmann:
Coming back to the use case at hand... @phackstock and I will try to implement a proper solution until the end of February. in the meantime, there are the following options:
 - do all aggregations on your side and submit the entire results at the `EU27 & EFTA` - which would then be just considered a native-region from the point of view of the ECEMF/nomenclature scenario processing
 - let `EU27 & EFTA` be aggregated automatically as part of the scenario submission, with the results reported at the EU-level being omitted for now
 - continue with the PR, where the two types of results are stored in "separate" regions

HauHe:
Hi @danielhuppmann ,
thanks for the plan and providing the suggestion on how to proceed. 
I think I will stick with option 2 for the moment and leave out the results reported at EU level for the moment, but will discuss it with the team here.

danielhuppmann:
Ok, sticking with option 2 means that no changes to the mappings and region definitions are necessary, so I'm closing this PR. Please re-open (or open a new one) if the situation changes. Please also follow the corresponding issue in the nomenclature repository so that you get an automated notificaton when we implemented a feature to support your use case.

PR #153, opened 2022-02-11
Added WITCH 5.1 (based on 34 regions) aggregation

JohannesEmm:
Adding region definition and mapping for WITCH.

I see quite some heterogeneity in what mappings are customary reported by the different models. I assume now with the automatic mapping, any desired region values can be computed correct, so not important that all mappings are manually entered (else there are quite a few missing across all models)?


mappings/witch_5.1.yaml
```diff
@@ -0,0 +1,83 @@
...
+  - ITA: WITCH 5.1|ITA
+  - NORTHEU: WITCH 5.1|NORTHEU
+  - POL: WITCH 5.1|POL
+  - PRT: WITCH 5.1|PRT
+  - ROU: WITCH 5.1|ROU
+  - SWE: WITCH 5.1|SWE
+  - EU27 & UK
+  - EU27
+  - World
+ common_regions:
```
danielhuppmann, line 40: There is an extra whitespace here which prevents parsing the yaml file correctly (this is why the GitHub Actions test fails)
```suggestion
common_regions:
```

definitions/region/model_native_regions/witch_v5.1.yaml
```diff
@@ -0,0 +1,38 @@
...
+  - WITCH 5.1|GBR
+  - WITCH 5.1|GRC
+  - WITCH 5.1|IRL
+  - WITCH 5.1|ITA
+  - WITCH 5.1|NORTHEU
+  - WITCH 5.1|POL
+  - WITCH 5.1|PRT
+  - WITCH 5.1|ROU
+  - WITCH 5.1|SWE
+
```
danielhuppmann, line 37: Please use exactly one empty line at the end of each file

definitions/region/model_native_regions/witch_v5.1.yaml
```diff
@@ -0,0 +1,38 @@
+# The model documentation for WITCH can be found at doc.witchmodel.org
+- WITCH 5.1:
+  - WITCH 5.1|BRAZIL
```
danielhuppmann, line 3: This is the region name how it will be displayed in the Scenario Explorer - you could change it to `WITCH 5.1|Brazil` for easier readability if you want...

danielhuppmann:
Not quite sure why the Actions is still failing even though the underlying issue has been resolved via #156 - will merge and hope that tests then pass...

PR #154, opened 2022-02-11
Adding weights for carbon, FE and SE prices

Renato-Rodrigues:


danielhuppmann:
@Renato-Rodrigues 

> Do I need to define those variables independently?

Yes, please add the newly created variables to the `definitions/variable` directory. Two reason why we opted against simply "over-riding" the list of allowed variables when they are generated by the **RegionProcessor**:
- We usually want to store additional annotation like definitions, units, etc., and the expected place for this are the definitions. It could be confusing to users if they do not find that information in the expected location.
- One can imagine a workflow where we want to run the **RegionProcessor**, then do additional, project-specific operations, and then do the validation. If a variable computed by the RegionProcessor is not in the list of allowed variables, this would fail.

Renato-Rodrigues:
do I need to add something like the below line to the aggregated variables declarations (`Price|Carbon (weighted by Final Energy)` and `Price|Carbon (weighted by Electricity Final Energy)`)?
```
skip-region-aggregation: true
```

danielhuppmann:
> do I need to add something like the below line to the aggregated variables declarations (`Price|Carbon (weighted by Final Energy)` and `Price|Carbon (weighted by Electricity Final Energy)`)?
> 
> ```
> skip-region-aggregation: true
> ```

Good question! I think it doesn't hurt to add, and you can then also add an attribute:
```yaml
    skip-region-aggregation: true
    note: This variable is computed as part of the region-processing from `Price|Carbon`.
```
 Or you could insert this explanation as a comment:
```yaml
# this variable is computed as part of the region-processing from `Price|Carbon`
skip-region-aggregation: true
```

Renato-Rodrigues:
Is there any way to rerun the Pytest without making a new commit?

definitions/variable/economy/economy.yaml
```diff
@@ -184,241 +184,263 @@
...
+        - Price|Carbon (weighted by Final Energy):
+            weight: Final Energy
+        - Price|Carbon (weighted by Electricity Final Energy):
+            weight: Final Energy|Electricity
+
+- Price|Carbon (weighted by Final Energy):
+    definition: Price of carbon
+    notes: Spatial aggregation use final energy as weighting variable. This is weight method
+     is preferable to avoid misleading aggregations when emissions are close to zero or negative.
+     This variable is computed as part of the region-processing from `Price|Carbon`.
```
danielhuppmann, line 202: The proposed note is grammatically not correct and also hard to understand for non-experts.
```suggestion
    notes: Using CO2 emissions as weights for average prices can cause spurious results
      when emissions are negative. This variable can be used as an alternative indicator
      for carbon prices which is less biased in low-carbon scenarios.
      This variable is computed as part of the region-processing from `Price|Carbon`.
```

danielhuppmann (commented):
See a few comments inline.

danielhuppmann (approved):
Thanks @Renato-Rodrigues

PR #158, opened 2022-03-02
Full country names in OSeMBE upload natives

HauHe:
This PR changes the country names in the mapping of the upload natives for OSeMBE to full country names that originate from the pyam mapping.


danielhuppmann (approved):
Looks good, thanks!

For clarification (I hope): the mapping per se has nothing to do with pyam...

- **pyam** - a general-purpose package for scenario processing
- **nomenclature** - a package for structured configuration of codelists and region mappings (using pyam to perform operations)
- this repository - the actual configuration used in a project

HauHe:
Yes, you are right. The country names come of course from the iso_mapping function in the openentrance package.

PR #159, opened 2022-03-07
Add valid ECEMF scenario names

Renato-Rodrigues:
adding file to define accepted ECEMF scenario names

### DIFF
--- a/definitions/scenario/scenarios.yaml
+++ b/definitions/scenario/scenarios.yaml
@@ -24,7 +24,7 @@
 - DIAG-C400-lin-ResidualFossil
 - DIAG-C400-lin-HighEff
 
-# ECEMF - Working Package 2 scenarios
+# ECEMF - Work package 2
 - WP2 Boundary Condition|Initial
 - WP2 Sensitivity|High Electricity Price
 - WP2 Sensitivity|High Gas and Liquids Price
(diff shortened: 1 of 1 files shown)


definitions/scenario/scenarios.yaml
```diff
@@ -0,0 +1,32 @@
+# Valid scenario names
+
+# ECEMF - Working Package 1 scenarios
```
danielhuppmann, line 3:
```suggestion
# ECEMF - Work package 1
```

definitions/scenario/scenarios.yaml
```diff
@@ -0,0 +1,32 @@
...
+- DIAG-C400-LimNuclear
+
+# Paradigm Shift scenarios
+- DIAG-C400-lin-HighVRE
+- DIAG-C400-lin-HighElectrification
+- DIAG-C400-lin-HighH2
+- DIAG-C400-lin-ResidualFossil
+- DIAG-C400-lin-HighEff
+
+# ECEMF - Working Package 2 scenarios
```
danielhuppmann, line 27:
```suggestion
# ECEMF - Work package 2
```

danielhuppmann (commented):
Suggest to use Work package instead of "Working package", otherwise good!

danielhuppmann (approved):
Thanks! Note that it will require a second PR until the scenario name validation will happen as part of the upload to the ECEMF-internal Scenario Explorer.

PR #161, opened 2022-03-09
IMAGE region support

MarkMDekker:
@danielhuppmann I think I resolved your comments. I followed your directions in renaming the mapping of IMAGE CEU + WEU towards "Europe (excl. Turkey)", while it is not perfect, as Ukraine, for example, is also not part of these IMAGE regions, while [this file](https://github.com/openENTRANCE/openentrance/blob/main/definitions/region/european-regions.yaml) seems to suggest it is part of "Europe (excl. Turkey)".


danielhuppmann (approved):
Looking good, many thanks!

danielhuppmann:
Note that you can also include any timeseries data at the regional level "Europe (excl. Turkey)" in your submission to the scenario data, and those values will be merged (and override) any values computed from the automated region-processing.

MarkMDekker:
Good to know - thanks!

PR #163, opened 2022-03-25
adding NAVIGATE variables emission variables

Renato-Rodrigues:
This pull request adds support for NAVIGATE variables from the categories:
* carbon management
* emissions (non-CO2)
* emissions (CO2)


definitions/variable/tag_fe_carriers.yaml
```diff
@@ -0,0 +1,95 @@
+# This list of inputs will replace any occurrence of `{Final Energy carrier}`
+# in a variable-definition yaml file
+
+- Final Energy carrier:
+  - Electricity:
+      description: electricity (including on-site solar PV), excluding transmission/distribution losses
+      weight: Electricity
```
phackstock, line 7: In analogy to my comment in #162 in the current definition `weight` is lost when applying the tags.
Adding `weight: "{Final Energy carrier}"` (you need the quotation marks otherwise yaml thinks you're defining a dictionary) to the variables where the tag is to be applied should fix it in case they are meant to get the appropriate `weight` attribute.

phackstock (commented):
Looks mostly fine to me.
Please find one comment below.

PR #164, opened 2022-03-28
Update accepted scenario list

Renato-Rodrigues:
- Correcting technology constraint scenario naming convention.

### DIFF
--- a/definitions/scenario/scenarios.yaml
+++ b/definitions/scenario/scenarios.yaml
@@ -13,9 +13,9 @@
 - DIAG-NZero
 
 # Technology Constraint scenarios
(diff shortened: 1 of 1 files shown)


danielhuppmann:
@Renato-Rodrigues, are you sure that you want to replace these scenario names? This PR means that previously submitted scenarios with those names will still exist in the database, but it will not be possible to update them.

Renato-Rodrigues:
This was a request from @robertpietzcker and it was discussed in the WP1 ECEMF meeting.
The scenario naming convention was updated for consistence, and all modeling teams already know that they will need to re-upload their scenario results using the new scenario naming convention.

PR #165, opened 2022-03-28
adding NAVIGATE services variables

Renato-Rodrigues:
add support for NAVIGATE services variables

### DIFF
--- a/definitions/variable/tag_transport_types.yaml
+++ b/definitions/variable/tag_transport_types.yaml
@@ -17,8 +17,9 @@
   - Maritime|Hydrogen:
       description: maritime hydrogen-based transportation technologies
 
-  - Rail:
-      description: rail-based transportation technologies
+# already in {transport mode}
+#  - Rail:
+#      description: rail-based transportation technologies
 
   - Rail|Electric:
       description: rail-based transportation technologies using electricity
--- a/definitions/variable/energy/energy-services.yaml
+++ b/definitions/variable/energy/energy-services.yaml
@@ -1,19 +1,44 @@
 # List of variables related to energy services
 
+- Energy Service|Transportation|Passenger:
+    description: Provision of energy services related to all passenger transportation modes
+    unit: Gpkm/yr
+- Energy Service|Transportation|Passenger|Bicycling and Walking:
+    description: Provision of energy services related to passenger transport on bicycles and by foot
+    unit: Gpkm/yr
+- Energy Service|Transportation|Passenger|{Transport mode}:
+    description: Provision of energy services related to passenger {Transport mode}
+    unit: Gpkm/yr
+- Energy Service|Transportation|Bunkers|{Bunkers mode}:
+    description: Provision of energy services related to passenger bunkers {Bunkers mode}
+    unit: Gpkm/yr
+- Energy Service|Transportation|Freight:
+    description: Provision of energy services related to all freight transportation modes
+    unit: Gtkm/yr
+- Energy Service|Transportation|Freight|{Transport mode}:
+    description: Provision of energy services related to freight {Transport mode}
+    unit: Gtkm/yr
+- Energy Service|Transportation|Freight|{Bunkers mode}:
+    description: Provision of energy services related to freight bunkers {Bunkers mode}
+    unit: Gtkm/yr
+
+- Energy Service|Residential and Commercial|Floor Space:
+    description: Provision of energy services for conditioned floor space in residential and commercial buildings
+    unit: bn m2/yr
+- Energy Service|Residential|Floor Space:
+    description: Provision of energy services for conditioned floor space in residential buildings
+    unit: bn m2/yr
+- Energy Service|Commercial|Floor Space:
+    description: Provision of energy services for conditioned floor space in commercial buildings
+    unit: bn m2/yr
+
+# additional variables
 - Energy Service|Transportation:
     description: Provision of energy services related to all transportation modes
     unit: Gvkm/yr
-- Energy Service|Transportation|Freight:
-    description: Provision of energy services related to all freight transportation
-      modes
-    unit: Gtkm/yr
 - Energy Service|Transportation|Freight|{Transport}:
     description: Provision of energy services related to freight {Transport}
     unit: Gtkm/yr
-- Energy Service|Transportation|Passenger:
-    description: Provision of energy services related to all passenger transportation
-      modes
-    unit: Gpkm/yr
 - Energy Service|Transportation|Passenger|{Transport}:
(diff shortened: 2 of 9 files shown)


Renato-Rodrigues:
Is there an additional check specific made to the {Transport} tag?
I had to disable the variable `Energy Service|Transportation|Freight|Rail` from the {Transport} tag to avoid duplicates, and the pull request got an error.

tests/test_definitions.py
```diff
@@ -32,14 +32,14 @@ def test_variables_industry_types():
...
 
-def test_variables_transport_types():
-    # check that exploding of {transport} to transportation modes works
-    obs = definition.variable["Energy Service|Transportation|Freight|Rail"]
-    exp = (
-        "Provision of energy services related to freight "
-        "rail-based transportation technologies"
-    )
-    assert obs["description"] == exp
+#def test_variables_transport_types():
```
danielhuppmann, line 35: Commenting out a test is only a good idea if you intend to re-activate it later. It is not clear to what this PR changed to make the test fail...

definitions/variable/technology/technologies.yaml
```diff
@@ -55,10 +63,18 @@
...
     skip-region-aggregation: true
+- Fixed Cost|Transportation|Freight|Rail:
+    description: Capital cost for freight rail-based transportation technologies
+    unit: [EUR_2020/tkm/yr, USD_2010/tkm/yr]
+    skip-region-aggregation: true
 - Fixed Cost|Transportation|Passenger|{Transport}:
     description: Capital cost for passenger {Transport}
     unit: [EUR_2020/pkm/yr, USD_2010/pkm/yr]
     skip-region-aggregation: true
+- Fixed Cost|Transportation|Passenger|Rail:
```
danielhuppmann, line 74: Why is this not part of the {Transport} tag-list?

definitions/variable/tag_transport_types.yaml
```diff
@@ -17,8 +17,8 @@
   - Maritime|Hydrogen:
       description: maritime hydrogen-based transportation technologies
 
-  - Rail:
-      description: rail-based transportation technologies
+#  - Rail:
```
danielhuppmann, line 20: Commenting out sections of code is usually not a good idea...

danielhuppmann (changes requested):
There are a few issues here...

tests/test_definitions.py
Renato-Rodrigues: As I mentioned above, Freight|Rail is not handled anymore by the {Transport} tag. To avoid a duplicated variable name error, the entry had to be removed from the {Transport} tag, which in turn caused this test to error out by default.

I commented it out only because I had no idea how you would like to proceed about that. 

The {Transport} tag became a legacy tag, and it was only left in the code due to the requirement to keep supporting non NAVIGATE variables.
In order to represent the NAVIGATE template variable it was necessary to redo the transport tag system in order support the new structure in transportation modes, technologies and carriers.

Renato-Rodrigues: Rail is not handled anymore by the {Transport} tag. To avoid a duplicated variable name error, the entry had to be removed from the {Transport} tag, which in turn caused this test to error out by default.

I commented it out only because I had no idea how you would like to proceed about the test: either remove it (1), or adapt it to the new NAVIGATE tags (tag_transport_mode.yaml, tag_transport_mode_and_carrier.yaml and tag_transport_mode_tech_and_carrier.yaml).

The {Transport} tag became a legacy tag, and it was only left in the code due to the requirement to keep supporting non NAVIGATE variables.
In order to represent the NAVIGATE template variable it was necessary to redo the transport tag system to support the new structure in transportation modes, technologies and carriers.

definitions/variable/tag_transport_types.yaml
Renato-Rodrigues: This is related with the previous points. This variable was commented to avoid duplication, but if I removed it directly it would appear that it was missing by mistake from the Transport tag.

Just let me know how would you prefer to proceed about it.

tests/test_definitions.py
phackstock: Coming is here as Daniel is on holiday, please excuse me as I might get this wrong.
As far as I read the test, it does not actually matter what the tag is called in the end `{Transport}` or in the newer case `{Transport Mode}`.
From what I gather, the test fails because you slightly changed the description from "rail-based transportation technologies" to "rail-based transportation". As the test is for an exact match, it fails.
You should be able to fix it by either changing `exp` the test to end with "rail-based transportation" or changing the description of `Rail` to "rail-based transportation technologies". 
You should also then be able to remove the former definition of `Rail` from `tag_transport_types.yaml`.

definitions/variable/tag_transport_types.yaml
phackstock: This connects to my comment about the commented out test case. `- Rail` should be safe to remove.

definitions/variable/technology/technologies.yaml
phackstock: This is where the second option you described in #162 would come in handy.
For now it's good to keep as is here.
Maybe open an issue to revisit this if or when that feature gets implemented.

phackstock (changes requested):
Looks good to me overall.
The "manual" mention of `Rail` is perfectly fine for now as there exists currently no mechanic in `nomenclature` that would allow to be selective about the application of tags.
The test should be relatively easy to adjust as detailed below. 
After that I'd say we're good to merge.

tests/test_definitions.py
Renato-Rodrigues: I am also not sure what was the intention behind the initial test and I did not dig deeper on checking its inner workings. 
I must say that after 5 years not using it, my python skills are a bit rust :) 

I simply commented it because it fails every time, so just let me know if I should simply delete the test code in the pull request, or if one of you could do a code commit suggestion on what should be changed to adapt the test to the new situation.

Renato-Rodrigues: handled in commit (https://github.com/openENTRANCE/openentrance/pull/165/commits/d249541f68daf901c990a93be901c34767f434cf)

definitions/variable/tag_transport_types.yaml
Renato-Rodrigues: handled in commit https://github.com/openENTRANCE/openentrance/pull/165/commits/a76334dc36d9fa33bb4f9f5dcb36f673daf8318a

phackstock:
Thanks for removing the comments and getting the test working again @Renato-Rodrigues.
Looks good from my side so I'll go ahead with the merge.

PR #166, opened 2022-03-28
adding NAVIGATE price variables

Renato-Rodrigues:
Is it possible to support the tag system in the weight definition?

e.g.
```
- Price|Final Energy|Industry|{Final Energy carrier}:
    description: price at the final level in the industrial sector of {Final Energy carrier}. 
    notes: Prices should include the effect of carbon prices and all other taxes.
    unit: [EUR_2020/GJ, USD_2010/GJ, EUR_2020/kWh]
    weight: Final Energy|Industry|{Final Energy carrier}
```


danielhuppmann:
> Is it possible to support the tag system in the weight definition?

Yes, this should work, but you have to provide a "weight"-attribute in the tag-dictionary, so that the tag can be replaced in the weight attribute of the actual codelist.

Renato-Rodrigues:
I am not sure I follow this. As the tag {Final Energy carrier} is used in multiple variables, if I provide a weight in its tag definition wouldn't all variables that use this tag to extend their definitions also be forced to use the same weight?
Was this considered in the implementation?
If this is the case, I will need to create a new tag only applicable to FE|Price variables copying all values from the current {Final Energy carrier} tag, so I can only apply the weights to the price variables and not to the other variables that use this tag.

definitions/variable/energy/energy-prices.yaml
```diff
@@ -0,0 +1,112 @@
...
+    unit: [EUR_2020/GJ, USD_2010/GJ, EUR_2020/kWh]
+    weight: Final Energy|Industry|{Final Energy carrier}
+
+- Price|Final Energy|Residential|{Final Energy carrier}:
+    description: price at the final level in the residential sector of {Final Energy carrier}. 
+    notes: Prices should include the effect of carbon prices and all other taxes.
+    unit: [EUR_2020/GJ, USD_2010/GJ, EUR_2020/kWh]
+    weight: Final Energy|Industry|{Final Energy carrier}
+
+- Price|Final Energy|commercial|{Final Energy carrier}:
```
danielhuppmann, line 60: This seems to be a typo
```suggestion
- Price|Final Energy|Commercial|{Final Energy carrier}:
```

definitions/variable/energy/energy-prices.yaml
```diff
@@ -0,0 +1,112 @@
...
+    description: price at the final level in the commercial sector of {Final Energy carrier}. 
+    notes: Prices should include the effect of carbon prices and all other taxes.
+    unit: [EUR_2020/GJ, USD_2010/GJ, EUR_2020/kWh]
+    weight: Final Energy|Industry|{Final Energy carrier}
```
danielhuppmann, line 64: This seems to be a copy-paste error...
```suggestion
    weight: Final Energy|Commercial|{Final Energy carrier}
```

definitions/variable/energy/energy-prices.yaml
```diff
@@ -0,0 +1,112 @@
...
+    description: price at the final level in the transport sector of {Final Energy carrier}. 
+    notes: Prices should include the effect of carbon prices and all other taxes.
+    unit: [EUR_2020/GJ, USD_2010/GJ, EUR_2020/kWh]
+    weight: Final Energy|Industry|{Final Energy carrier}
+
+- Price|Final Energy|Residential and Commercial|{Final Energy carrier}:
+    description: price at the final level in the residential and commercial sector of {Final Energy carrier}. 
+    notes: Prices should include the effect of carbon prices and all other taxes.
+    unit: [EUR_2020/GJ, USD_2010/GJ, EUR_2020/kWh]
+    weight: Final Energy|Industry|{Final Energy carrier}
```
danielhuppmann, line 52: This seems to be a copy-paste error
```suggestion
    weight: Final Energy|Residential and Commercial|{Final Energy carrier}
```

definitions/variable/energy/energy-prices.yaml
```diff
@@ -0,0 +1,112 @@
...
+    description: price at the final level in the industrial sector of {Final Energy carrier}. 
+    notes: Prices should include the effect of carbon prices and all other taxes.
+    unit: [EUR_2020/GJ, USD_2010/GJ, EUR_2020/kWh]
+    weight: Final Energy|Industry|{Final Energy carrier}
+
+- Price|Final Energy|Transportation|{Final Energy carrier}:
+    description: price at the final level in the transport sector of {Final Energy carrier}. 
+    notes: Prices should include the effect of carbon prices and all other taxes.
+    unit: [EUR_2020/GJ, USD_2010/GJ, EUR_2020/kWh]
+    weight: Final Energy|Industry|{Final Energy carrier}
```
danielhuppmann, line 46: This seems to be a copy-paste error
```suggestion
    weight: Final Energy|Transportation|{Final Energy carrier}
```

definitions/variable/tag_fe_carriers.yaml
```diff
@@ -0,0 +1,95 @@
...
+
+  - Waste:
+      description: non-renewable waste
+      note: only for industry and non-energy use sectors
+      weight: Waste
+      
+  - Other:
+      description: other sources that do not fit to any other category
+      weight: Other
+
```
danielhuppmann, line 94: Please have exactly one empty line at the end of each file.

definitions/variable/energy/energy-prices.yaml
```diff
@@ -0,0 +1,112 @@
...
+
+- Price|Final Energy|commercial|{Final Energy carrier}:
+    description: price at the final level in the commercial sector of {Final Energy carrier}. 
+    notes: Prices should include the effect of carbon prices and all other taxes.
+    unit: [EUR_2020/GJ, USD_2010/GJ, EUR_2020/kWh]
+    weight: Final Energy|Industry|{Final Energy carrier}
+
+# additional variables    
+- Price|Secondary Energy|Gases|Natural Gas:
```
danielhuppmann, line 67: It would make sense to also add this as a tag-list in this PR...

danielhuppmann (changes requested):
There are several copy-paste errors and a few issues...

definitions/variable/energy/energy-prices.yaml
Renato-Rodrigues: This variable was kept only due to backwards compatibility reasons in the db. 
After some clarification with Robert, my understanding is that the "Natural Gas" keyword should be replaced by "Gas" for consistence in the NAVIGATE template, and secondary energy prices should follow the structure of final energy carriers. 
Therefore, the variable "Price|Secondary Energy|Gases|Fossil" should already reflect the natural gas prices at secondary energy level as far as I understand, and natural gas specific variables would make more sense at the primary energy level.

Nevertheless, if both of you think otherwise (@danielhuppmann and @robertpietzcker ), I can add this additional split to the {Final Energy carrier} tag, although I think it is unecessary.

robertpietzcker: To my understanding, having the SE-prices on the disaggregation level fossil/electricity/biomass is sufficient for most questions at hand, so adding the prices for all the subcomponents of fossil (Gas/Oil/Coal for the energy carriers gases, liquids, electricity, heat and hydrogen) will create many new variables without adding much insights at the moment. 

So I would go with Renato's proposal to not move that to the full tagging system.

I am not sure how much backwards compatibility is needed. In principle, I would be in favor of dropping these variables, as they simply don't fit the new system anymore.  If backwards compatibility is needed on the SE level, I would propose to just keep the three idiosyncratic variables "Price|Secondary Energy|Gases|Natural Gas", "Price|Secondary Energy|Solids|Coal" "Price|Secondary Energy|Liquids|Oil".
(although it pains me a bit to keep this one instance of "Natural Gas" when we decided to remove the "Natural" from all other variables and just stick to "Gas")

robertpietzcker (approved):

definitions/variable/energy/energy-prices.yaml
phackstock: From the scenario explorer point of view it is not necessary to keep this variable as these definitions only affect data during upload. So if there's no new data coming with this variable it would be fine to be removed.
For bookkeeping purposes we might want to keep it though but maybe expand the `notes` attribute to make it clear that this variable is no longer to be used.

phackstock (approved):
Looks good to me, ready to be merged.
Very cool that you're pushing the nomenclature tag functionality to be used in the weight attribute. 
I have added a comment to the discussion about `Price|Secondary Energy|Gases|Natural Gas`.

Renato-Rodrigues:
Daniel asked me to keep backwards compatibility because the repository could be in use by other projects, but if we reach an agreement that we could remove these legacy variables it would make life much easier, we could further extend the tags use in the future, and it would make the variables definition files much cleaner and concise.
Maybe we should get back to this discussion when he is back.

phackstock:
Ah right, I see, then let's maybe create an issue here to revisit the discussion. If it's to be used by multiple projects then we should indeed be careful about removing variables.
In this case this PR is good to be merged from my side.

PR #173, opened 2022-05-04
Update energy-final.yaml

sandrinecharousset:
correct tag for Final Energy|electricity|profile
Add variable share|final energy|electricity|cooling: this variable is necessary for performing case studies with plan4EU using the openentrance scenario which do not separate the cooling part of elec consumption. This % is then taken eg from the ehighway dataset or from litterature


danielhuppmann (approved):
Note that we currently have an issue saving dimensionless units (like a share) to the ixmp database infrastructure. Will need to investigate and resolve that.

PR #176, opened 2022-05-04
Update power-plant.yaml

sandrinecharousset:
updated variables with regard to openentrance CS1
added list of energy storage systems


danielhuppmann (approved):
Than

PR #181, opened 2022-05-11
Update electricity-grid.yaml

sandrinecharousset:
add variable network|electricity|length
@loeffko

### DIFF
--- a/definitions/variable/technology/electricity-grid.yaml
+++ b/definitions/variable/technology/electricity-grid.yaml
@@ -3,6 +3,12 @@
 - Network|Electricity|Length:
     description: length of a line
     unit: Mkm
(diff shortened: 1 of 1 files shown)


definitions/variable/technology/electricity-grid.yaml
```diff
@@ -1,5 +1,14 @@
 # List of variables related to the electricity grid
 
+- Network|Electricity|Length:
+    description: length of a line
+    unit: Mkm
+- Network|Electricity|Expansion Cost:
+    description: Cost for increasing the line capacity
+    unit: MEUR/GW/km
```
danielhuppmann, line 8: I suggest to avoid "MEUR", please use "million EUR/GW/km" or "EUR/KWh/km"

PR #182, opened 2022-05-11
Update technologies.yaml

sandrinecharousset:
added variables: Variable Cost, Variable Cost with/without emission cost for electricity inputs

### DIFF
--- a/definitions/variable/technology/technologies.yaml
+++ b/definitions/variable/technology/technologies.yaml
@@ -95,6 +95,26 @@
     description: Variable cost for {Electricity Input}, without the emission cost
     unit: [EUR_2020/MWh, USD_2010/MWh]
     skip-region-aggregation: true
+- Levelized Cost|Capex|Electricity|{Electricity Input}:
+    description: Levelized cost for investment in {Electricity Input}
+    unit: [MEUR_2020/GW, MUSD_2010/GW, MEUR_2020/PJ, MUSD_2010/PJ]
+    skip-region-aggregation: true
+- Levelized Cost|Generation|Electricity|{Electricity Input}:
+    description: Levelized cost for generation of electricity by {Electricity Input}
+    unit: [MEUR_2020/MWh, MUSD_2010/MWh, MEUR_2020/PJ, MUSD_2010/PJ]
+    skip-region-aggregation: true
+- Levelized Cost|Emissions|Electricity|{Electricity Input}:
+    description: Levelized cost for investment in {Electricity Input}
+    unit: [MEUR_2020/MWh, MUSD_2010/MWh, MEUR_2020/PJ, MUSD_2010/PJ]
+    skip-region-aggregation: true
+- Levelized Cost|Total (incl. Emissions)|Electricity|{Electricity Input}:
+    description: Levelized cost for investment in {Electricity Input}
+    unit: [MEUR_2020/MWh, MUSD_2010/MWh, MEUR_2020/PJ, MUSD_2010/PJ]
+    skip-region-aggregation: true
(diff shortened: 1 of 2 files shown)


sandrinecharousset:
@loeffko

definitions/variable/technology/technologies.yaml
```diff
@@ -76,13 +76,45 @@
...
-    description: Capital cost for {Fuel}
+    description: Variable cost for {Fuel}
     unit: [EUR_2020/MWh, USD_2010/MWh]
     skip-region-aggregation: true
 - Variable Cost|Heat|Industrial|{Fuel}:
-    description: Capital cost for {Fuel}
+    description: Variable cost for {Fuel}
+    unit: [EUR_2020/MWh, USD_2010/MWh]
+    skip-region-aggregation: true
+- Variable Cost (incl. Fuel Cost)|Electricity|{Electricity Input}:
```
danielhuppmann, line 86: I suggest to instead interpret "Variable Cost" as *only* the non-fuel-related costs and then add a separate variable "Fuel Cost|Electricity|..."

definitions/variable/technology/technologies.yaml
```diff
@@ -76,13 +76,45 @@
...
+    description: Variable cost for {Electricity Input}, including the fuel cost
+    unit: [EUR_2020/MWh, USD_2010/MWh]
+    skip-region-aggregation: true
+- Variable Cost (excl. Fuel Cost)|Electricity|{Electricity Input}:
+    description: Variable cost for {Electricity Input}, without the fuel cost
     unit: [EUR_2020/MWh, USD_2010/MWh]
     skip-region-aggregation: true
+- Levelized Cost|Capex|Electricity|{Electricity Input}:
```
danielhuppmann, line 94: Is this really different from the existing variable "Capital Cost|Electricity|..."?

danielhuppmann (commented):
A few questions and suggestions inline.

definitions/variable/technology/technologies.yaml
sandrinecharousset: Well, I am also going to add the variable 'Fuel Cost' but the way of getting the Variable Cost (incl. Fuel Cost) is not at all trivial and this is an output of Genesys-Mod so I would like to keep the Variable Cost (incl. Fuel Cost) as sth that can be retrieved from the platform as it is a required input for some models.
But we can remove the Variable Cost (excl. Fuel Cost) and add in the description of 'Variable Cost' that it is excluding the Fuel cost (which is not trivial and was previously mis-interpreted by more than one modelling team)

loeffko: I agree with Daniel (and this is also how we have the variable costs natively in our model). So in my understanding, the variable costs are just those associated with the technology itself.

I can provide them separately without an issue, so adding a separate variable sounds good to me!

definitions/variable/technology/technologies.yaml
loeffko: Yes, this is indeed different, as @sandrinecharousset says. For capital costs, we list the actual investment costs in M€/GW, whereas within the levelized cost calculations for the technology, they are essentially converted to a "per energy produced" basis.

sandrinecharousset: thanks @loeffko for the 2 comments
@danielhuppmann seems we are all ok now, so this PR can be merged, right?

loeffko (approved):

sandrinecharousset:
Closing this PR as some of the changes were already included and the remaining one (levelized costs) is in a new PR

PR #184, opened 2022-05-11
Update tag_electricity_input_types.yaml

sandrinecharousset:
added different kinds of gas power plant (per gas type, power plant type, and CCS or not)
added different types of PV power plants (rooftops, utilities...)

### DIFF
--- a/definitions/variable/tag_electricity_input_types.yaml
+++ b/definitions/variable/tag_electricity_input_types.yaml
@@ -103,6 +103,39 @@
   - Gas|Fossil|w/o CCS:
       description: fossil natural gas (excluding methane from biomass or hydrogenation)
         with freely vented CO2 emissions
+  - Gas|Fossil|OCGT:
+      description: fossil natural gas (excluding methane from biomass or hydrogenation)
+        in an open-cycle gas turbine (OCGT) power plant 
+  - Gas|Fossil|CCGT:
+      description: fossil natural gas (excluding methane from biomass or hydrogenation)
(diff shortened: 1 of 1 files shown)


definitions/variable/tag_electricity_input_types.yaml
```diff
@@ -103,14 +103,80 @@
   - Gas|Fossil|w/o CCS:
       description: fossil natural gas (excluding methane from biomass or hydrogenation)
         with freely vented CO2 emissions
+  - Gas|Fossil|OCGT:
+      description: fossil natural gas (excluding methane from biomass or hydrogenation)
+        in an open-cycle gas turbine (OCGT) power plant 
+  - Gas|Fossil|CCGT:
+      description: fossil natural gas (excluding methane from biomass or hydrogenation)
+        in a combined-cycle gas turbine (CCGT) power plant
+  - Gas|Fossil|Gas Engines:
```
danielhuppmann, line 112: I know of "Gas Turbines" - do you mean those?

danielhuppmann (commented):
See the clarifying question below.

definitions/variable/tag_electricity_input_types.yaml
sandrinecharousset: No seems to be different
This name was discussed with Konstantin as it is what is used in Genesys-Mod and it seems to be used (konstantin showed me a publication by siemens explaining the difference between gas turbines and gas engines
I sent you the email

PR #189, opened 2022-05-19
Add variables for space heating/cooling and water heating

danielhuppmann:
As requested by @ozerece and @andreasm2 for ECEMF WP2.2.


danielhuppmann:
After some discussion with @ozerece and Gustav Resch, we agreed that Ambient Heat should not be included as an "energy carrier" in the sense of the final-energy classification. I will start an issue to discuss alternative approaches to represent the relevant information about heat-pump efficiency and energy streams.

PR #191, opened 2022-05-31
A minus sign intead of an empty value in variables within the electricity-grid.yaml file 

erikfilias:
This PR proposes the usage of the minus sign instead of an empty value as a unit on the following variables:
```
Network|Electricity|Fixed Charge Rate:
Network|Electricity|Loss Factor:
Network|Electricity|Reactance:
Network|Electricity|Security Factor:
```


danielhuppmann (approved):
Thanks @erikfilias

PR #194, opened 2022-06-23
Add Energy Storage System Type to possible electricity capacities

loeffko:
This introduces the option to give installed capacities for storages, as requested in the openENTRANCE case studies.
This is different from the Maximum Storage (capacity vs energy content).

### DIFF
--- a/definitions/variable/technology/technologies.yaml
+++ b/definitions/variable/technology/technologies.yaml
@@ -4,6 +4,9 @@
 - Capacity|Electricity|{Electricity Input}:
     description: Total installed electricity generation capacity from {Electricity Input}
     unit: GW
+- Capacity|Electricity|{Energy Storage System Type}:
+    description: Total installed electricity generation capacity from {Energy Storage System Type}
+    unit: GW
 - Capacity|Heat|Residential and Commercial:
     description: Total installed heat generation capacity for the residential and
       commercial sectors
--- a/definitions/variable/technology/technologies.yaml
+++ b/definitions/variable/technology/technologies.yaml
@@ -5,7 +5,7 @@
     description: Total installed electricity generation capacity from {Electricity Input}
     unit: GW
 - Capacity|Electricity|Energy Storage System:
-    description: Total installed electricity generation capacity from all {Energy Storage System Type} storage systems
+    description: Total installed electricity generation capacity from all storage systems
     unit: GW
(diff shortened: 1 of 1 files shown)


definitions/variable/technology/technologies.yaml
```diff
@@ -4,6 +4,9 @@
 - Capacity|Electricity|{Electricity Input}:
     description: Total installed electricity generation capacity from {Electricity Input}
     unit: GW
+- Capacity|Electricity|{Energy Storage System Type}:
```
sandrinecharousset, line 7:
```suggestion
- Capacity|Electricity|Energy Storage System:
    description: Total installed electricity generation capacity from all {Energy Storage System Type} storage systems
    unit: GW
- Capacity|Electricity|Energy Storage System|{Energy Storage System Type}:
```
For consistency with the other variables related to energy storage, and also use of the aggregations
I also propose to include the aggregated capacity of all batteries

sandrinecharousset (changes requested):
For consistency with other variables related to energy storage systems, I suggest to afdd 'Energy Storage System before the tag, and also to add the variable with the aggregated capacity

definitions/variable/technology/technologies.yaml
loeffko: Sounds good to me - potentially, I would name it 
Capacity|Electricity|Energy Storages
since that sounds nicer in my opinion. But if this is the naming used everywhere else, then we can of course use this.

Also, yes, it would then also provide an aggregated number of all electricity storages.

sandrinecharousset: @loeffko I would have agreed for Energy Storages but the other name is used everywhere so.....

loeffko: @sandrinecharousset done. Should be fine now.

definitions/variable/technology/technologies.yaml
```diff
@@ -4,6 +4,12 @@
 - Capacity|Electricity|{Electricity Input}:
     description: Total installed electricity generation capacity from {Electricity Input}
     unit: GW
+- Capacity|Electricity|Energy Storage System:
+    description: Total installed electricity generation capacity from all {Energy Storage System Type} storage systems
```
danielhuppmann, line 8: If the tag is not used in the variable name, it does not make sense to use it in the description
```suggestion
    description: Total installed electricity generation capacity from all storage systems
```

PR #199, opened 2022-06-25
Update electricity-operation.yaml

sandrinecharousset:
Add variables Import/Export/Non Served Demand

### DIFF
--- a/definitions/variable/technology/electricity-operation.yaml
+++ b/definitions/variable/technology/electricity-operation.yaml
@@ -10,7 +10,8 @@
 #   - Demand in reserves
 #   - Demand in inertia
 #   - Flows in transmission lines
-# Demand not served
+# 
+
 
 - Active Power|Electricity|{Electricity Input}:
     description: Active power of {Electricity Input} power plants
@@ -91,3 +92,15 @@
     description: Marginal Cost associated to the Frequency Containment requirement
     unit: EUR_2020/MWh
     skip-region-aggregation: true
+- Export|Electricity:
+    description: Exported electricity from a given region
+    unit:MWh
+    skip-region-aggregation: true 
+- Import|Electricity:
+    description: Exported electricity from a given region
+    unit:MWh
+    skip-region-aggregation: true 
+- Non Served Demand|Electricity:
+    description: Not served demand
+    unit:MWh
+    skip-region-aggregation: true 
--- a/definitions/variable/technology/electricity-operation.yaml
+++ b/definitions/variable/technology/electricity-operation.yaml
@@ -102,7 +102,7 @@
 - Unserved Demand|Electricity:
     description: Not served demand
     unit: MWh
-- Charged Energy|Electricity|{Electricity Input}:
+- Injected Energy|Electricity|{Electricity Storage Type}:
     description: Amount of energy charged to the storage during the period
     notes: this represents the total energy which is charged into the storage during the period, the energy which is discharged during 
      that same period is not substracted from this total. This means that the Net Charged Energy can be computed as Charged Energy - Discharged Energy
--- a/definitions/variable/technology/electricity-operation.yaml
+++ b/definitions/variable/technology/electricity-operation.yaml
@@ -95,12 +95,12 @@
 - Export|Electricity:
     description: Exported electricity from a given region
     unit:MWh
-    skip-region-aggregation: true 
+    skip-region-aggregation: true
 - Import|Electricity:
     description: Exported electricity from a given region
     unit:MWh
-    skip-region-aggregation: true 
-- Non Served Demand|Electricity:
+    skip-region-aggregation: true
+- Non served Demand|Electricity:
     description: Not served demand
     unit:MWh
-    skip-region-aggregation: true 
+    skip-region-aggregation: true
--- a/definitions/variable/technology/electricity-operation.yaml
+++ b/definitions/variable/technology/electricity-operation.yaml
@@ -104,9 +104,14 @@
     unit: MWh
 - Charged Energy|Electricity|{Electricity Input}:
     description: Amount of energy charged to the storage during the period
+    notes: this represents the total energy which is charged into the storage during the period, the energy which is discharged during 
+     that same period is not substracted from this total. This means that the Net Charged Energy can be computed as Charged Energy - Discharged Energy
+     if Charged Energy > Discharged Energy, 0 if not
     unit: GWh
 - Discharged Energy|Electricity|{Electricity Input}:
     description: Amount of energy charged to the storage during the period
-    notes: This has to be a positive value. If more energy is charged into storage than discharged from storage , please used the variable "Charged Energy"
+    notes: this represents the total energy which is discharged from the storage during the period, the energy which is charged into the storage during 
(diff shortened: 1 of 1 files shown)


definitions/variable/technology/electricity-operation.yaml
```diff
@@ -91,3 +92,19 @@
...
     unit: EUR_2020/MWh
     skip-region-aggregation: true
+    
+- Export|Electricity:
+    description: Exported electricity from a given region
+    unit: MWh
+- Import|Electricity:
+    description: Exported electricity from a given region
+    unit: MWh
+- Non served Demand|Electricity:
```
danielhuppmann, line 102: I'd rather call this variable "Unserved Demand" or "Curtailed Demand"

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -91,3 +92,19 @@
...
+    description: Not served demand
+    unit: MWh
+- Charged Energy|Electricity|{Electricity Input}:
+    description: Amount of energy charged to the storage during the period
```
danielhuppmann, line 106: If you think having two separate variables is the right approach, I suggest to add the following note to avoid confusion between injection (charging) and extraction (discharging).

```suggestion
    description: Amount of energy charged to the storage during the period
    notes: This has to be a positive value. If more energy is discharged from storage than charged into storage, please used the variable "Discharged Energy"
```

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -91,3 +92,19 @@
...
+    unit: GWh
+- Discharged Energy|Electricity|{Electricity Input}:
+    description: Amount of energy charged to the storage during the period
```
danielhuppmann, line 109:
```suggestion
    description: Amount of energy charged to the storage during the period
    notes: This has to be a positive value. If more energy is charged into storage than discharged from storage , please used the variable "Charged Energy"

```

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -91,3 +92,19 @@
...
+- Export|Electricity:
+    description: Exported electricity from a given region
+    unit: MWh
+- Import|Electricity:
+    description: Exported electricity from a given region
+    unit: MWh
+- Non served Demand|Electricity:
+    description: Not served demand
+    unit: MWh
+- Charged Energy|Electricity|{Electricity Input}:
```
danielhuppmann, line 105: I'm wondering if "Charged Energy" is sufficiently intuitive, and the term also doesn't quite work for other erngy carriers than electricty. Maybe use "Net Injection into Storage" (and use a negative value if more energy is discharged than charged)?

danielhuppmann (commented):
A few comments inline

definitions/variable/technology/electricity-operation.yaml
erikfilias: I suggest the name `Not-Served Demand|Electricity`. I also agree with the "Unserved Demand".

definitions/variable/technology/electricity-operation.yaml
erikfilias: For me, "Charged Energy" is intuitive but would be associated with the new storage type tag when it is merged.

erikfilias (approved):

definitions/variable/technology/electricity-operation.yaml
sandrinecharousset: @danielhuppmann Curtailed demand does not fit as it is sth different: it refers in general to voluntary energy curtailments, while non served is missing energy
I will change to Unserved
@erikfilias I don't much like the '-' in the variable name

definitions/variable/technology/electricity-operation.yaml
sandrinecharousset: @danielhuppmann @erikfilias the idea here is that the 2 variables can be positive at the same time, eg if you have 3 time steps in the period covered by the variable and energy is discharged in the 1st time step and charged in the other timesteps, then the Charged energy will be the sum of the 2 last timesteps while Discharged Energy will be the 1st timestep. I add a comment to explain that in the variable. 
@erikfilias agree?

danielhuppmann (changes requested):
I'm confused by the variable `Charged Energy|Electricity|{Electricity Input}` - it is usually not possible to know "which" electricity goes into storage (same as import/export). I suggest to simplify to `Charged Energy|Electricity` and `Discharged Energy|Electricity`.

definitions/variable/technology/electricity-operation.yaml
erikfilias: I also agree. In [openTEPES](https://opentepes.readthedocs.io/en/latest/MathematicalFormulation.html), we use the same concept of having two variables.

definitions/variable/technology/electricity-operation.yaml
sandrinecharousset: I also agree with @erikfilias , but Net Injection into Storage is even better , maybe could be simplified into Injected Energy? (then we would have Injected Energy and Discharged Energy?

sandrinecharousset:
> I'm confused by the variable `Charged Energy|Electricity|{Electricity Input}` - it is usually not possible to know "which" electricity goes into storage (same as import/export). I suggest to simplify to `Charged Energy|Electricity` and `Discharged Energy|Electricity`.

@danielhuppmann I do not agree. Yes in the real life you cannot know which electricity goes into whch storage, but you can know how much electricity is injected/discharged to/from any storage. Plus, this variable is a model output and we need for results analyses to know how each kind of storage is used. So we do need these variables and not only aggregates

definitions/variable/technology/electricity-operation.yaml
danielhuppmann: This suggestion implements the name change above, and it changes the tag-list from the energy-input-fuels to the storage types (so that we can track to which storage type energy is injected/discharged).
```suggestion
- Injected Energy|Electricity|{Electricity Storage Type}:
```

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -91,3 +92,26 @@
...
+- Unserved Demand|Electricity:
+    description: Not served demand
+    unit: MWh
+- Charged Energy|Electricity|{Electricity Input}:
+    description: Amount of energy charged to the storage during the period
+    notes: this represents the total energy which is charged into the storage during the period, the energy which is discharged during 
+     that same period is not substracted from this total. This means that the Net Charged Energy can be computed as Charged Energy - Discharged Energy
+     if Charged Energy > Discharged Energy, 0 if not
+    unit: GWh
+- Discharged Energy|Electricity|{Electricity Input}:
```
danielhuppmann, line 111: Change of tag-list as above.
```suggestion
- Discharged Energy|Electricity|{Electricity Storage Type}:
```

definitions/variable/technology/electricity-operation.yaml
sandrinecharousset: ok , Injected sounds better :-)

danielhuppmann (approved):
Thanks @sandrinecharousset!

PR #196, opened 2022-06-24
Creation of the tag electricy storage types

erikfilias:
This PR adds the tag `tag_electricity_storage_types` that comprises all technology able to store energy (Hydro reservoir, PSH, ESS, etc.).
Then, this PR also proposes some modifications:

- to remove Pumping Efficiency / Turbine Efficiency
- to keep Roundtrip Efficiency
- to add Charge Efficiency and Discharge Efficiency (which would be equivalent to Technology2Storage / TechnologyFromStorage from genesis-mod. We are aware this is the same in genesis mod, but not the case with other sources of data)

closes #183

PR #201, opened 2022-06-26
Update energy-secondary.yaml

sandrinecharousset:
Add secondary energy|electricity|electricity storage type
COmplement to #196

### DIFF
--- a/definitions/variable/energy/energy-secondary.yaml
+++ b/definitions/variable/energy/energy-secondary.yaml
@@ -11,6 +11,9 @@
 - Secondary Energy|Electricity|{Electricity Input}:
     description: Net electricity production from {Electricity Input}
     unit: EJ/yr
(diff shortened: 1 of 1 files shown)


loeffko (commented):
Is net electricity production the right term in this case?
Storages do not really produce anything, even less "net" - where they would only have losses.

Maybe something like "Electricity discharge from {Electricity Storage Type} might be more fitting?

sandrinecharousset:
@loeffko agreed

sandrinecharousset:
I propose to delete this PR as , as mentioned by @loeffko , 'storages', except for Hydro, do not produce electricity. Hydro is already icnluded in the variables names via {Electricity Input}, and we don't want a variable such as Secondary Energy|Electricity|Lithium-Ion

PR #207, opened 2022-07-27
Add extra models from the remind group

Renato-Rodrigues:
This PR intends to add support for additional models from the REMIND group that share the same the same region definition.

**Question:** @danielhuppmann @phackstock 
How to proceed with the native regions definition?
Right now we expressly write the `REMIND 2.1` string in the native regions definition so we keep the uploaded data saved under the model name. However, as all models included in this pull request share the same regional definition, it would be much better if the model name part of the tag `REMIND 2.1|<region name>` would be created automatically by the db code, instead of creating multiple almost identical region mapping files, except for the replaced model name tag in the file.
Is this feature possible to be implemented? 

**Suggestion:**
Instead of using 
```
  - CAZ: REMIND 2.1|CAZ
```
support a key word like for example:
```
  - CAZ: <model>|CAZ
```
that would be filled automatically at the data processing time with the model name.

Please let me know if this would be possible or if I would need to go with the multiple files solution, with repeated regions definitions.

### DIFF
--- a/mappings/remind_2.1.yaml
+++ b/mappings/remind_2.1.yaml
@@ -1,4 +1,9 @@
-model: REMIND 2.1
+model: 
+  - REMIND 2.1
+  - REMIND 3.0.0
+  - REMIND-Industry 3.0.0
+  - REMIND-Buildings 3.0.0
+  - EDGE-T 0.15.0
 native_regions:
   - CAZ: REMIND 2.1|CAZ
   - CHA: REMIND 2.1|CHA


phackstock:
Thanks for the suggestion @Renato-Rodrigues. In my opinion it should not be too difficult to implement such a feature. With the tags for the definitions we already have something similar. It makes total sense to me to have such a feature as we introduced the ability to have the same region mapping for multiple models. 
I would ask you to open an issue on https://github.com/iamconsortium/nomenclature when we can see who is going to take a closer look.

danielhuppmann:
Thanks for the suggestion @Renato-Rodrigues. We might implement this feature at some point in the future, but for now, please
- start a new PR with two new mapping files using the model-identifier is „REMIND 3.0“ and „EDGE 0.15” respectively
- close this PR with a reference to the new PR

Renato-Rodrigues:
closing this PR as suggested and creating an alternative one that can be found in: https://github.com/openENTRANCE/openentrance/pull/208

PR #208, opened 2022-07-27
Add extra models from the remind group using additional files

Renato-Rodrigues:
as suggested in: https://github.com/openENTRANCE/openentrance/pull/207


mappings/REMIND-Buildings_v3.0.0.yaml
```diff
@@ -0,0 +1,57 @@
+model: REMIND-Buildings 3.0.0
```
danielhuppmann, line 1:
```suggestion
model:
  - REMIND-Buildings 3.0.0
  - REMIND-Industry 3.0.0
```

mappings/REMIND-Buildings_v3.0.0.yaml
```diff
@@ -0,0 +1,57 @@
...
+native_regions:
+  - CAZ: REMIND-Buildings 3.0.0|CAZ
```
danielhuppmann, line 3:
```suggestion
  - CAZ: REMIND 3.0|CAZ
```

definitions/region/model_native_regions/REMIND-Buildings_v3.0.0.yaml
```diff
@@ -0,0 +1,23 @@
+# REMIND regions mapping can be found at https://git.io/JySgR and https://doi.org/10.1016/j.energy.2021.121908
+- REMIND-Buildings 3.0.0:
+  - REMIND-Buildings 3.0.0|CAZ
```
danielhuppmann, line 3:
```suggestion
- REMIND 3.0:
  - REMIND 3.0|CAZ
```

danielhuppmann (changes requested):
To reduce the number of region-groups, I suggest to add regions for the hierarchy "REMIND 3.0", which are then used for the submodels Industry and Buildings and for all patch-versions 3.0.x. I implemented suggestions to clarify.



(same for "EDGE-T 0.15"

mappings/REMIND-Buildings_v3.0.0.yaml
Renato-Rodrigues: I understand your suggestion, but we need to differentiate these model runs from pure REMIND 3.0.0 run results that should be added soon to the db. So I would keep instead the different names and the PR in its original form.

danielhuppmann: But does REMIND 3.0.0 have a different region resolution? If not, I'd use the same region resolution.

Renato-Rodrigues: REMIND 3.0.0 has the same regional resolution. 
At the same time I would like to keep in the db the results from REMIND 2.1, REMIND 3.0.0, REMIND-Buildings 3.0.0 and REMIND-Industry 3.0.0 differentiated in the db so we can compare their results against each other.

Renato-Rodrigues: I just didn't included yet the REMIND 3.0.0 model to the db as its results are still not available, and if in the meantime you have the opportunity to implement the feature to use a single region mapping file, I would prefer instead to refactor these files to simplify this when I do the REMIND 3.0.0 results upload.

danielhuppmann: Of course the results from the three models will be kept separate in the database. My comment just referred to the geographical resolution - if all three models use the same regions, that should be made explicit. Plus it allows easier filtering and visualization of the data.

Renato-Rodrigues: Thks, I think I understand what do you mean now. 
In that case I would say I don't even need these extra files, no? 
I could simply change the originals `REMIND 2.1` region mapping files (`mappings/remind_2.1.yaml` and `definitions/region/model_native_regions/REMIND_2.1.yaml`) to use something like:
`- CAZ: REMIND|CAZ` 
and declare all models in a single `mappings/remind.yaml` file instead

Renato-Rodrigues: What about the `definitions/region/model_native_regions` file?
Could I simply create a single file named `remind.yaml` that include: 
```
REMIND:
  - CAZ: REMIND|CAZ
  - ...
```
or do I need to keep one for each model?

danielhuppmann: I would keep the region-mapping specific for the MAJOR.MINOR model version because the regional resolution may change in the future - and CAZ may be a different region in REMIND 4.0 compared to REMIND 2.1 and 3.0…
It’s all about finding a good compromise between having too many categories vs not enough categories.

Renato-Rodrigues: This was deliberated internally already and we have planed to keep compatibility in between the past and future region acronomys used in REMIND. So this shouldn't be an issue in our case.

danielhuppmann: I find it very hard to believe that there will not be any changes in the regional resolution of the REMIND model over the next years. Please keep the region-mapping such that the region-hierarchy includes the MAJOR.MINOR model version.

Renato-Rodrigues: There will be changes in the regional resolution of REMIND, but they will always be incremental.
Including the version number in the region definition is just misleading in this case, as when we reach REMIND 4 with more regions, instead of only adding the extra regions to the region definition, I will have to edit the version number unnecessarily in more than one file.
If this is a mandatory format from the iiasa db I will change it, although I don't see the point of including the version number in something that will be version independent.

danielhuppmann: There are a few issues in your response, but the discussion goes beyond this PR. Please follow my advice on versioning of the region-resolution.

mappings/remind.yaml
```diff
@@ -0,0 +1,62 @@
+model: 
+  - REMIND 2.1
```
danielhuppmann, line 2: Not sure why you want to use REMIND 3.0-regions for REMIND 2.1? Better to just leave that as it was...

Renato-Rodrigues: As I said before, REMIND region definitions are agnostic in relation to the model version.
The REMIND region acronyms are unique and they will be always possible to be reported in any current or future version of REMIND.
I changed my PR to replace the 3.0 version by the 2.1 as requested if you prefer to keep that way, although I still can't see any reason that would justify to include any version number to it.

danielhuppmann:
closing in favor of #210

PR #210, opened 2022-07-29
Add region and mapping for REMIND 3.0

danielhuppmann:
This PR implements the discussion from #208 about REMIND 3.0 regions as separate files keeping the existing model-mapping for REMIND 2.1 unchanged (to not alter behavior for existing model submissions to the IIASA database infrastructure).

We will also add guidance on region-naming conventions, see https://github.com/IAMconsortium/nomenclature/issues/142


Renato-Rodrigues (approved):

danielhuppmann:
Permissions to upload to the ECEMF-internal Scenario Explorer instance granted to all REMIND modelers.

PR #211, opened 2022-08-08
Update tag_electricity_input_types.yaml

sandrinecharousset:
Add CHP and non CHP plants

### DIFF
--- a/definitions/variable/tag_electricity_input_types.yaml
+++ b/definitions/variable/tag_electricity_input_types.yaml
@@ -58,6 +58,15 @@
       description: lignite coal and sub-bituminous coal with a CO2 capture component
   - Coal|Lignite|w/o CCS:
       description: lignite coal and sub-bituminous coal with freely vented CO2 emissions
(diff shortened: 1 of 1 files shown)


danielhuppmann:
For clarification - do we want to to keep track of the combination of CHP and CCS plants? i.e., have a variable to keep track how much electricity was generated in a CHP that also has a carbon-capture module...

sandrinecharousset:
> For clarification - do we want to to keep track of the combination of CHP and CCS plants? i.e., have a variable to keep track how much electricity was generated in a CHP that also has a carbon-capture module...

@danielhuppmann , it is not needed for the current case studies of openEntrance (CHP is needed in CS4), so I did not include it as I did not know if there was a specific rule for the order (w/ CHP|w/o CCS or w/o CCS|w/CHP , but this may be needed in the future; I could add it

danielhuppmann:
Ok, thanks for the clarification, let's add it when we actually need it...

PR #213, opened 2022-08-10
Remove HeatPump

sandrinecharousset:
HeatPumps are not geterating electricity (but consuming electricity ..... and generating heat/cold....) thus should not be included in this tag
Extend description of the file in the header


danielhuppmann (approved):
Thanks @sandrinecharousset

PR #215, opened 2022-08-19
Add pumped-hydro as storage option

danielhuppmann:
This PR follows on #196, related to #199.

The PR clarifies the distinction between "new" electricity generation at pumped-hydro power plants (from natural water inflow) versus the injection/discharge at these plants (as storage operation).

An alternative would be to include all "new" electricity generation  as "Hydro|Reservoir".

### DIFF
--- a/definitions/variable/tag_electricity_input_types.yaml
+++ b/definitions/variable/tag_electricity_input_types.yaml
@@ -284,7 +284,10 @@
   - Hydro|Pumped Storage:
       description: hydropower systems with short-term storage capacity
         (including pumping turbines)
-      notes: This hydropower system has two water reservoirs at different elevations.
+      notes: This variable should only include electricity generation from natural
+        inflow of water into the upper reservoir. Electricity from storage operation
+        should be denoted by *"Injected Energy|.."* and *Discharded Energy|..*.
+        This hydropower system has two water reservoirs at different elevations.
         The definition follows the convention established
         by the ENTSO-E transparency platform (https://transparency.entsoe.eu).
   - Hydro|Reservoir:
--- a/definitions/variable/tag_electricity_input_types.yaml
+++ b/definitions/variable/tag_electricity_input_types.yaml
@@ -284,10 +284,7 @@
   - Hydro|Pumped Storage:
       description: hydropower systems with short-term storage capacity
         (including pumping turbines)
-      notes: This variable should only include electricity generation from natural
-        inflow of water into the upper reservoir. Electricity from storage operation
-        should be denoted by *"Injected Energy|.."* and *Discharded Energy|..*.
-        This hydropower system has two water reservoirs at different elevations.
+      notes: This hydropower system has two water reservoirs at different elevations.
         The definition follows the convention established
         by the ENTSO-E transparency platform (https://transparency.entsoe.eu).
   - Hydro|Reservoir:
--- a/definitions/variable/tag_electricity_storage_types.yaml
+++ b/definitions/variable/tag_electricity_storage_types.yaml
@@ -6,12 +6,12 @@
 - Electricity Storage Type:
 
   - Energy Storage System:
-      description: Mainly to refer to the batteries
+      description: all types of energy storage including pumped hydro and batteries
   - Energy Storage System|Pumped Hydro Storage:
       description: pumped-hydro storage
   - Energy Storage System|Lithium-Ion:
-      description: Lithium-Ion batteries
+      description: lithium-ion batteries
   - Energy Storage System|Redox Flow:
-      description: Redox Flow batteries
+      description: redox flow batteries
   - Energy Storage System|Compressed Air:
-      description: Compressed Air batteries
+      description: compressed air storage systems
--- a/definitions/variable/tag_electricity_storage_types.yaml
+++ b/definitions/variable/tag_electricity_storage_types.yaml
@@ -7,6 +7,8 @@
 
   - Energy Storage System:
       description: Mainly to refer to the batteries
+  - Energy Storage System|Pumped Hydro Storage:
+      description: pumped-hydro storage
   - Energy Storage System|Lithium-Ion:
       description: Lithium-Ion batteries
   - Energy Storage System|Redox Flow:
--- a/definitions/variable/tag_electricity_storage_types.yaml
+++ b/definitions/variable/tag_electricity_storage_types.yaml
@@ -23,8 +23,6 @@
         (possibly including pumping turbines)
       notes: This definition follows the convention established
         by the ENTSO-E transparency platform (https://transparency.entsoe.eu).
-  - Energy Storage System|Pumped Hydro Storage:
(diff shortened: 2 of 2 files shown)


definitions/variable/tag_electricity_storage_types.yaml
```diff
@@ -6,10 +6,12 @@
 - Electricity Storage Type:
 
   - Energy Storage System:
-      description: Mainly to refer to the batteries
+      description: all types of energy storage including pumped hydro and batteries
+  - Energy Storage System|Pumped Hydro Storage:
```
erikfilias, line 10: In my opinion, there are three main points:

- `Energy Storage System|Pumped Hydro Storage` should be removed since we already have `Hydro|Pumped Storage`.
- `Hydro|Reservoir`, `Hydro|Pumped Storage` and `Energy Storage System|*` have variables in common related to the *storage operation*.
- `Hydro|Pumped Storage` takes electricity from the grid to store water which is the main difference concerning `Energy Storage System|*`. This characteristic can be also applied to thermal energy storage for `Solar|CSP`.

So, perhaps we can create a new category for this kind of storage where the power unit takes electricity from the grid to store another energy source type that will be used to produce electricity later.

danielhuppmann: Please take another look at the current status and #199, in particular for which variables the tags `Electricity Storage Type` vs. `Electricity Input` are used.

Also, all storage types take electricity from the grid to store it as another source of energy (gravity, kinetic, chemical, ...) and then revert it back to electriciy.

erikfilias: Ok, I also agree with all storage types of energy and their variables injected/discharged.

So, I got the purpose of defining `Energy Storage System|Pumped Hydro Storage`. Since we can use the storage operation variables for it and the variables related to electricity operation by using `Hydro|Pumped Storage`. 
Both definitions for PHS are fine.

Moreover, we could perhaps define `Energy Storage System|Hydro Reservoir` in case the variables related to reservoir operation would be needed.

sandrinecharousset, line 10: if we want to include all kind of storages as defined above, we should also add 
Energy Storage System|Hydro|Reservoir, as this also is another kind of storage. It usually does not take electricity from the grid to store it in a specific form but it could (see below) and it is a way to store electricity.
In the modelling we are using, which is quite rough, Reservoir units do not have pumping assets, but in real life there exist some 'Reservoir' kind of units with also Pumping assets. The difference is mainly the size of the reservoir ('Reservoir' is for big size reservoirs, which are managed by electricity system operators on a yearly base, whlle 'Pumped Storage' units are of lower size and thus are managed more on a weekly or day/night base.
Anyway, in the present state of the needs, we do not need the specific 'storage system' variables for the 'Reservoir' so this could wait.......

sandrinecharousset, line 10: Why not call this Energy Storage System|Hydro|Pumped Storage? Would be more consistent with the way hydro is defined

definitions/variable/tag_electricity_input_types.yaml
```diff
@@ -284,7 +284,10 @@
   - Hydro|Pumped Storage:
       description: hydropower systems with short-term storage capacity
         (including pumping turbines)
-      notes: This hydropower system has two water reservoirs at different elevations.
+      notes: This variable should only include electricity generation from natural
```
sandrinecharousset, line 287: I don't understand this..... We cannot 'separate' electricity generated from the natural inflows and electricity generated from water which has previously been pumped........... A pumped hydro unit will generate electricity (xMW at each hour), and pump water while consuming electricity (yMW at each hour); Note that it is possible to 'pump' and 'turbine' simultaneously (and this has proven very efficient, in particular regarding system services, so numerous pumped storage units are currently being given this possibility); Also I do not understand why restrictions are detailed here about a 'variable' while this is only a 'tag' list.....

definitions/variable/tag_electricity_storage_types.yaml
sandrinecharousset: I also noticed that in 'Electricity input' it is mentioned that 'Reservoir' can have pumps

sandrinecharousset (changes requested):

definitions/variable/tag_electricity_storage_types.yaml
danielhuppmann: The tag *Energy Storage Types* is used for the just-added variables for injected and discharged energy into storage. In my opinion, if it is natural inflow into a reservoir, this is not discharged energy (in a storage sense) but simply electricity generation optimized over time, hence just an electricity-generation input.

And yes, I see that there is a note in the definition of hydro-reservoir that that it can have pumps, but I do not understand why this was added there in the first place. I vaguely recall that you and/or @erikfilias wanted to insert that a while ago.

sandrinecharousset: @danielhuppmann I both agree and disagree. yes 'Reservoir' is exactly what you mention, in the way we are using it now (with the data from the scenarios...); But in the real life 'Reservoir' is acting alltogether as the 2 concepts. It is electricity generation optimized over time, and it also is (for 'Reservoir' with pumps) discharged energy in storage sense. So if we want to be strictly following the definition, we should add it as a storage option (maybe with a note telling it is the case ONLY for reservoirs with pumping capacity); 
But now we do not need that (at least for the case studies that I know), we are not using any data with a detailed modelling of hydro such that it includes the disaggrgated reservoirs with also their pumping facilities (and in the main detailed modellings that are used for mid to long term horizons, even when disaggregated, pumping capacities are usually treated separately from 'reservoir' capacities). In the case we would like to run case studies with detailed 'daily like' operation optimisation then we would need to include reservoirs with pumping. 
But as I already said, we can keep that for a later stage and not bother with it now.

definitions/variable/tag_electricity_input_types.yaml
sandrinecharousset: to complement the above, a Pumped Hydro system could also have natural inflows, thus could both generate additionnal electricity AND 'discharge ' stored electricity -(ir generate electricity from water which was previously pumped); My model allows that (but it is not currently used as we do not have data for inflows to pumped hydro storages.....)

definitions/variable/tag_electricity_storage_types.yaml
danielhuppmann: Updated as suggested for consistency

definitions/variable/tag_electricity_input_types.yaml
danielhuppmann: I removed my attempt at disambiguation.

danielhuppmann:
I'm not sure how the distinction between pumped-hydro vs. reservoir as storage should be implemented in practice by modelers, but I have followed your requests and implemented both as a storage-type.

PR #216, opened 2022-08-19
Electrification supply additional scenario

Renato-Rodrigues:
- scenario "DIAG-C400-lin-HighElec-Supply" added to the scenarios list for ECEMF


danielhuppmann (approved):
Thanks @Renato-Rodrigues!

PR #223, opened 2022-08-24
Re-insert additional prices at the secondary-energy level

### LINKED ISSUE #221: Secondary energy prices removed from the variables definition
Could I ask why secondary energy prices were removed from the variables definition files?

https://github.com/openENTRANCE/openentrance/commit/4730be9b4133f0f74e5c798e866f4e4477acd5aa#diff-9720cdd30a1b714d520da007c71b80ea771e9856553f248aaabd08d84f377adeL28

Ex: It can be quite important to know the price levels for fossil, biomass and/or synfuel based gases at secondary energy level. The same for liquids, solids and, specially, hydrogen. However, the above linked commit removed them from the db. 
The less than handful entries that remained in the database do not provide enough information to deeper analysis. 
Besides, the removed variables were already being in use for ECEMF analysis carried in WP1 and they are part of the variables template used by the modeling teams.

I would kindly request to the removed variables to be reinstated the database.

danielhuppmann:
This PR adds variables for prices at the secondary energy level. The variables match those used in the ENGAGE project.

closes  #221

### DIFF
--- a/definitions/variable/energy/energy-prices.yaml
+++ b/definitions/variable/energy/energy-prices.yaml
@@ -69,12 +69,12 @@
     unit: [EUR_2020/GJ, USD_2010/GJ]
     weight: Secondary Energy|Gases
 
-- Price|Secondary Energy|Gases|Natural Gas:
+- Price|Secondary Energy|Gases|Fossil:
     description: Natural gas price at the secondary level, i.e. for large-scale
       consumers (e.g., gas power plant)
     notes: Prices should include the effect of carbon prices.
     unit: [EUR_2020/GJ, USD_2010/GJ]
-    weight: Secondary Energy|Gases|Natural Gas
+    weight: Secondary Energy|Gases|Fossil
 
 - Price|Secondary Energy|Gases|Biomass:
     description: Biogas price at the secondary level, i.e. for large-scale consumers
@@ -108,6 +108,11 @@
     notes: Prices should include the effect of carbon prices.
     unit: [EUR_2020/GJ, USD_2010/GJ]
     weight: Secondary Energy|Solids|Biomass
+
+- Price|Secondary Energy|Solids|Biomass|Traditional:
+    description: Price for traditional biomass at the secondary level
+    unit: [EUR_2020/GJ, USD_2010/GJ]
+    weight: Secondary Energy|Solids|Biomass|Traditional
 
 - Price|Secondary Energy|Hydrogen:
     description: Hydrogen price at the secondary level, i.e. for large-scale consumers
--- a/definitions/variable/energy/energy-prices.yaml
+++ b/definitions/variable/energy/energy-prices.yaml
@@ -62,6 +62,13 @@
     unit: [EUR_2020/GJ, USD_2010/GJ]
     weight: Secondary Energy|Electricity
 
+- Price|Secondary Energy|Gases:
+    description: Price for gaseous fuels at the secondary level, i.e. for large-scale
+      consumers (e.g., gas power plant)
+    notes: Prices should include the effect of carbon prices.
+    unit: [EUR_2020/GJ, USD_2010/GJ]
+    weight: Secondary Energy|Gases
+
 - Price|Secondary Energy|Gases|Natural Gas:
     description: Natural gas price at the secondary level, i.e. for large-scale
       consumers (e.g., gas power plant)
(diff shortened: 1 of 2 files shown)


danielhuppmann:
@Renato-Rodrigues, please let me know if you think we need additional price variables at the secondary level...

Renato-Rodrigues:
I am afraid there would be still some missing variables. 
The full list of secondary energy prices included in the ECEMF variables template is copied below for reference:

```
Price|Secondary Energy|Electricity
Price|Secondary Energy|Gases
Price|Secondary Energy|Gases|Biomass
Price|Secondary Energy|Gases|Electricity
Price|Secondary Energy|Gases|Fossil
Price|Secondary Energy|Heat
Price|Secondary Energy|Hydrogen
Price|Secondary Energy|Hydrogen|Biomass
Price|Secondary Energy|Hydrogen|Electricity
Price|Secondary Energy|Hydrogen|Fossil
Price|Secondary Energy|Liquids
Price|Secondary Energy|Liquids|Biomass
Price|Secondary Energy|Liquids|Electricity
Price|Secondary Energy|Liquids|Fossil
Price|Secondary Energy|Other
Price|Secondary Energy|Solar
Price|Secondary Energy|Solids
Price|Secondary Energy|Solids|Biomass
Price|Secondary Energy|Solids|Biomass|Traditional
Price|Secondary Energy|Solids|Fossil
```

The previous code responsible to create those variables used the tag system to simplify the variables creation. I copy below the single item responsible to create all those entries before, in case you want to restore the way these variables were created.

```
# Secondary energy prices:
- Price|Secondary Energy|{Final Energy carrier}:
    description: price at the secondary level of {Final Energy carrier}, i.e. for large scale
                consumers (e.g. aluminum production, gas power plant)
    notes: Prices should include the effect of carbon prices.
    unit: [EUR_2020/GJ, USD_2010/GJ]
    weight: Secondary Energy|{Final Energy carrier}
```

danielhuppmann:
The reason why I removed the automatically generated variables for the secondary-energy prices is that they where inconsistent with the Secondary-Energy-variables, so those could not be used for automated region-aggregation. I have now explicitly re-inserted most of your listed variables, except for:

- Price|Secondary Energy|Gases|Fossil
  -> identical as "Price|Secondary Energy|Gases|Natural Gas", possible confusion with stream-reformed hydrogen

- Price|Secondary Energy|Other
  -> this is not specific enough to be reported anyway

- Price|Secondary Energy|Solar
  -> this is not a thing

- Price|Secondary Energy|Solids|Biomass|Traditional
  -> traditional biomass is usually not traded and therefore does not have a (modelled) price

Renato-Rodrigues:
> * Price|Secondary Energy|Gases|Fossil
>   -> identical as "Price|Secondary Energy|Gases|Natural Gas", possible confusion with stream-reformed hydrogen

Wouldn't it be more consistent to keep the same naming convention that is used in all other variables? 
We also report the "Secondary Energy|Gases|Fossil" variable, so it makes more sense to me to keep using the same naming convention for the corresponding prices. Besides I don't see possibility for confusion also, as steam-reforming hydrogen production would be reported under SE|Hydrogen|Fossil, as it is producing hydrogen and not gases, no?

> * Price|Secondary Energy|Other
>   -> this is not specific enough to be reported anyway

I don't have a strong opinion about this, although I think that including the "other" field allow more results to be reported by models that only consider aggregates and/or do not follow exactly the same disaggregation that IAMs traditionally do for energy carriers. We need to keep in mind that using results from bottom-up models require more flexibility than the models that we are used to.    

> * Price|Secondary Energy|Solar
>   -> this is not a thing

nothing to add.

> * Price|Secondary Energy|Solids|Biomass|Traditional
>   -> traditional biomass is usually not traded and therefore does not have a (modelled) price

Prices can represent the marginal costs of producing traditional biomass in this case. This can be important to the analysis to compare costs in models that consider only traditional biomass technics, versus models that have the differentiation.

In a general note, I don't understand why these variables were removed in the first place. The repository should be project specific, and when I did the variables update, you expressly said to me to keep backwards compatibility with the previous supported variables. It is much more trouble for the project members to have to get back and have a required renaming or removal of the current variables from outside of the project changes.

danielhuppmann:
I checked the secondary-energy-gas hierarchy and noticed that there are currently three variables to describe the same thing:

- Secondary Energy|Gases|Fossil
- Secondary Energy|Gases|Natural Gas
- Secondary Energy|Gases|Gas

There is also *Secondary Energy|Gases|Coal*, so *Gases from Fossil* is - technically speaking - the sum of gas and gas from coal, but I assume the latter is only a small fraction of the total.

This PR now also removes *Secondary Energy|Gases|Gas* to reduce ambiguity, but we may need to revisit at a later point in time. (Such inconsistencies or errors are also a reason why a template/codelist will never be final).

I have now changed the price-variable to *Price|Secondary Energy|Gases|Fossil* as requested, and I have added *Price|Secondary Energy|Solids|Biomass|Traditional*.

Adding *Other* makes sense if the aim is to make sure that totals add up (i.e., the energy mix), but I would prefer to wait until someone actually presents a use case for an *Other Price* variable...

Renato-Rodrigues (approved):

Renato-Rodrigues:
Just as an additional note, I would try to be consistent also on the use of either "Natural gas" or "Gas" in the variables naming convention.
Primary energy uses Gas in the naming convention, so I would say it is a much bigger change to enforce the natural gas change, than simply keep using "Gas" to reflect "natural gas".

danielhuppmann:
Ok, good point, I removed „Natural Gas“ and kept „Gas“.

PR #224, opened 2022-09-05
hot-fix for WP2 industry scenarios: adds EUR region for REMIND

mellamoSimon:
adds aggregated EUR region to REMIND's mapping


danielhuppmann:
Ignore the failing unit tests (that's a different issue, will fix this) - but I'm confused whether `EUR` is really a native region? I assume it's the aggregation of `DEU`, `FRA` and a few other regions...

Please describe which problem or issue you are trying to address with this PR.

mellamoSimon:
yess, sorry. 
`DEU`, `FRA`, etc. are native regions of REMIND when used in its 21-region configuration, which is what we want for ECEMF. I still think I can manage to have runs in that configuration on time for a ECEMF-WP2 meeting this week, in which case this change wouldn't be needed.
However, I might need to use REMIND runs in the 12-region configuration, where the native region is called `EUR`. In that case, I would need that region (`EUR`) to be mapped for this project.

danielhuppmann:
In my understanding, a model with a different regional disaggregation level should have a different version number, right? Or there should be some other identifier in the model name to distinguish the number of regions....

Renato-Rodrigues:
@mellamoSimon could you delete this PR?
The recommended way to upload REMIND results is described at this already merged PR: https://github.com/openENTRANCE/openentrance/pull/311 

EUR results should only be uploaded in H12 runs, and in EU21 runs they are calculated automatically by the iiasa db.

PR #225, opened 2022-09-07
Update message_1.2.yaml for "World" region

behnam-zakeri:
This PR adds the "World" region to the existing mapping file.


danielhuppmann (approved):
Thanks @behnam-zakeri!

PR #226, opened 2022-09-09
Updated IMAGE regions

MarkMDekker:
I added all IMAGE native regions here. Is this the correct syntax? I will in a minute also adapt the mapping file.


definitions/region/model_native_regions/image_v3.2.yaml
```diff
@@ -2,4 +2,28 @@
...
+  - IMAGE 3.2|RSAS
+  - IMAGE 3.2|RUS
+  - IMAGE 3.2|SAF
+  - IMAGE 3.2|SEAS
+  - IMAGE 3.2|STAN
+  - IMAGE 3.2|TUR
+  - IMAGE 3.2|UKR
+  - IMAGE 3.2|USA
+  - IMAGE 3.2|WAF
+  - IMAGE 3.2|World
```
danielhuppmann, line 29: There is no need a model-specific "World" region, you can just use the common "World" region for global data.
```suggestion

```

PR #228, opened 2022-09-09
Update IMAGE mapping try 2

MarkMDekker:


MarkMDekker:
@danielhuppmann do you see what I am doing wrong here? I tried to update both the native regions (other pull request) and the mapping between regions; include all IMAGE regions instead of only the European ones, and in particular the "World" region. Perhaps it needs some time to process the native regions, before changing the mapping?

danielhuppmann:
The validation checks that the region-mapping is consistent with the defined regions (in `definitions/region`). The validation is failing because #226 is not merged yet. Preferably (i.e., next time), you should add the region-definition and the mapping as one PR.

MarkMDekker:
Clear. Will combine the changes in one PR next time. Thanks for your quick response!

danielhuppmann (approved):
> I will provide all data as "World", following your first suggestion (Daniel). I guess these changes are in order then, without a specific common region?

This looks ok, @MarkMDekker - and the tests are passing now because the regions are now defined in the project specifications.

PR #231, opened 2022-10-05
Added EU27 to common regions

sirinalibas:


danielhuppmann (approved):
Thanks @sirinalibas, looks good!

PR #232, opened 2022-10-27
Add region-mappings for sectoral models of FORECAST

danielhuppmann:
Adding region-mappings for the sectoral modules of the FORECAST model, per email request by @sirinalibas @MatthiasRehfeldt

fyi @behnam-zakeri

### DIFF
--- a/mappings/FORECAST_v1.0.yaml
+++ b/mappings/FORECAST_v1.0.yaml
@@ -1,5 +1,7 @@
 model:
   - FORECAST v1.0
+  - FORECAST-Industry v1.0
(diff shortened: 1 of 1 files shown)


danielhuppmann:
Note that the GitHub Actions project validation is failing due to an issue in a downstream dependency (https://github.com/IAMconsortium/units/issues/40) which is not a problem for merging this PR.

danielhuppmann:
Permissions for upload the new models to the ECEMF-internal Scenario Explorer added for the group "Fraunhofer ISI Forecast Modelers".

PR #233, opened 2022-11-11
Simplify region-aggregation-mapping for OSeMBE

danielhuppmann:
Per a discussion with @HauHe, this PR simplifies the region-aggregation-mapping for OSeMBE to directly use country names rather than the `<mode>|<region>` structure for easier comparison across models.


danielhuppmann:
This PR is failing because of some ongoing work on the **nomenclature** package. This failing test does not affect the PR or the operation of the openENTRANCE and ECEMF Scenario Explorer instances.

HauHe:
Great thanks a lot Daniel!

PR #234, opened 2022-11-15
Change name to area_name

phackstock:
Changing the attribute `name` in the region codelist files `nuts1.yaml`, `nuts2.yaml` and `nuts3.yaml` to `area_name`. 
An attribute named `name` causes issues with the latest nomenclature release.
Running the tests, it seems we have another issue that will need to be fixed on the nomenclature side concerning multiple units.

### DIFF
--- a/definitions/region/nuts1.yaml
+++ b/definitions/region/nuts1.yaml
@@ -4,488 +4,488 @@
 # List of NUTS1 regions: major socio-economic regions
 - NUTS1:
   - BE1:
-      name: Région de Bruxelles-Capitale/Brussels Hoofdstedelijk Gewest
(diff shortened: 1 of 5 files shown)


danielhuppmann (approved):
Thank you, good to be merged once the other issue in nomenclature is fixed and the tests pass…

phackstock:
Looks like there's a few gremlins left. I'll investigate further.

phackstock:
Looks like there's a few last issues to fix. I'm taking a look.

phackstock:
Looks like it's all good now. @danielhuppmann, I'll go ahead with the merge to restore the service.

PR #236, opened 2022-11-23
Use country names directly instead of model-specific renaming

danielhuppmann:
Following a suggestion by @HauHe, this PR changes the region-mapping for Euro-Calliope to use country names directly instead of the more cumbersome `<model>|<country>` currently applied as part of the renaming upon submission to the ECEMF-internal Scenario Explorer. A similar change was made for the OSeMBE model in #233.

The idea is to change to a system where the model-specific region names are only used where the model does have non-obvious region names or aggregates. For models with a national resolution, we will directly use the country names without renaming.

This PR does not affect any scenario data already uploaded to the Scenario Explorer - it only changes the processing for future uploads. I will clean up existing data at a later stage.

For future uploads of Euro-Calliope, there are two minor changes required in your data:
- "Netherlands" -> "The Netherlands"
- "Northern Macedonia" -> "North Macedonia"

Please take a look @FLomb and @sjpfenninger!
fyi @phackstock


sjpfenninger:
Fine with me!

FLomb:
Sure, please go ahead!

PR #238, opened 2022-12-05
additional WP1 - task 1.3 scenarios

Renato-Rodrigues:


danielhuppmann:
Just for double-checking - other scenarios names have an identifier "WP2". Are we sure that these names do not need a WP identifier?

Renato-Rodrigues:
> Just for double-checking - other scenarios names have an identifier "WP2". Are we sure that these names do not need a WP identifier?

These are the scenario names defined in the WP1 scenarios protocol from ECEMF.
As far as I understand, they do not include the specific WP tag because the scenarios protocol is shared openly. This is the same case of the previously defined diagnosis scenario. 
The aim is to allow other models that do not participate in the project to create the same scenarios for comparison. Therefore, including the WP tag (e.g. WP1) would lose meaning in this context, as these scenarios use is not limited to the working package work only.

danielhuppmann:
ok, thanks for the clarification!

PR #241, opened 2022-12-09
Add LIMES v 2.38 mappings

joanna-sitarz:
Include file limes_v2.38.yaml in mappings.


danielhuppmann (approved):
Thanks, looks good to me!

I added permission for all REMIND-modelers to submit scenarios of the model "LIMES 2.38" to the ECEMF-internal Scenario Explorer instance.

PR #246, opened 2022-12-20
Remind's region mapping for MEESA

tatar99:
An extention of regions mapping in MEESA to improve comparability of regional results with other models.

### DIFF
--- a/definitions/region/model_native_regions/meesa_v1.1.yaml
+++ b/definitions/region/model_native_regions/meesa_v1.1.yaml
@@ -8,3 +8,10 @@
   - MEESA v1.1|DEU
   - MEESA v1.1|FRA
   - MEESA v1.1|POL
+  - MEESA v1.1|ECE
+  - MEESA v1.1|ECS
+  - MEESA v1.1|ENC
+  - MEESA v1.1|ESC
+  - MEESA v1.1|ESW
+  - MEESA v1.1|EWN
+  - MEESA v1.1|NEN 


danielhuppmann:
Thanks for the PR, but to better understand the purpose... ECE, ECS are not actually MEESA native regions, but are aggregates of native regions that match the REMIND regions?

If yes, there are two better approaches:
1. define ECE, ECS as common-regions and add a mapping to them in the MEESA and REMIND (and other) mappings)
2. (at least) make the mapping from MEESA-native regions to ECE, ECS by using the common-region feature in the mapping (and only upload actual native regions to the ECEMF-internal Scenario Explorer instance):
    ```yaml
    common_regions: 
      - MEESA v1.1|ECE:
          - MEESA v1.1|DEU
          - MEESA v1.1|FRA
          - ...
    ```
    This way, the mapping becomes more transparent (and you can upload smaller files without duplicating information).

tatar99:
Yes, your understanding is correct. 
Hauke needs our results for DIAG-400-lin scenario in aggregation similar to REMIND regions.
But I don't want to permanently change our native regions - it was intendent as a one-time action. 
That's why I thought that defining new common-regions for MEESA was not necessary.
Alternatively, perhaps, It would be more convenient to define a new MEESA version with this new aggregation?
What do you think?

danielhuppmann:
Pending discussion/agreement of #247

danielhuppmann:
Closing in favor of #248

PR #247, opened 2022-12-20
Add the INNOPATHS regions

danielhuppmann:
Per email conversations with @tatar99 and the PRIMES team, this PR adds the regions from the INNOPATHS project for easier comparison between REMIND, PRIMES, and other models.

I took the liberty to come up with "readable" region names instead of the abbreviations and added the suffix `(IP)` such that the regions can easily be identified in a large ensemble (with many different region aggregations, similar to adding `(R5)` in the global region).

FYI @HauHe

### DIFF
--- a/definitions/region/european-regions.yaml
+++ b/definitions/region/european-regions.yaml
@@ -102,3 +102,11 @@
   - United Kingdom & Ireland (IP):
       innopaths: UKI
       countries: UK, Ireland
+
+- European (non-EU) north countries (IP):
+      innopaths: NEN
+      countries: Iceland, Liechtenstein, Norway, Switzerland
+
+- European (non-EU) south countries (IP):
+      innopaths: NES
+      countries: Albania, Andorra, Bosnia and Herzegovina, North Macedonia, Monaco, Montenegro, Serbia, Turkey
--- a/definitions/region/european-regions.yaml
+++ b/definitions/region/european-regions.yaml
@@ -102,11 +102,10 @@
   - United Kingdom & Ireland (IP):
       innopaths: UKI
       countries: UK, Ireland
-
-- European (non-EU) north countries (IP):
+  - Northern (non-EU) Europe (IP):
       innopaths: NEN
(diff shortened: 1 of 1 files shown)


HauHe:
To me this looks good. If I'm not mistake it is the mapping that REMIND is already using, no?
@tatar99, I don't see NEN, but for my use that is ok, since NEN didn't contain any EU countries.

tatar99:
@HauHe Yes, I checked this is the same mapping we were taking about, except for NEN. Looks good to me too.

definitions/region/european-regions.yaml
```diff
@@ -72,3 +72,33 @@
...
+      countries: Portugal, Spain
+  - Central Europe (IP):
+      innopaths: EWN
+      countries: Austria, Belgium, Luxembourg, Netherlands
+  - France (IP):
+      innopaths: FRA
+      countries: France
+  - United Kingdom & Ireland (IP):
+      innopaths: UKI
+      countries: UK, Ireland
```
Renato-Rodrigues, line 104: For the sake of completeness I would add the following two extra regions.

```suggestion
      countries: UK, Ireland

- European (non-EU) north countries (IP):
      innopaths: NEN
      countries: Iceland, Liechtenstein, Norway, Switzerland

- European (non-EU) south countries (IP):
      innopaths: NES
      countries: Albania, Andorra, Bosnia and Herzegovina, North Macedonia, Monaco, Montenegro, Serbia, Turkey
```

Maybe you can come up with a better naming convention than this. 
The NEN region can be specially important as it includes non-eu countries that are part of the ETS (Iceland, Liechtenstein, Norway) or have an ETS system in place (Switzerland).

Renato-Rodrigues (changes requested):

definitions/region/european-regions.yaml
danielhuppmann: Thanks @Renato-Rodrigues!

PR #248, opened 2022-12-21
Add INNOPATHS mappings to MEESA

danielhuppmann:
@tatar99, this is the mapping if you want to include the regions *ECE*, *ECS*, ... etc in your uploads - they will be renamed to the INNOPATHS regions.

Alternatively, you could let the IIASA database infrastructure to the aggregation for by specifying the mapping, e.g.,
```yaml
  - Germany (IP):
      - DEU
  - Eastern Europe (IP):
      - CEU
      - POL
  ...
```

### DIFF
--- a/definitions/model/README.md
+++ b/definitions/model/README.md
@@ -1,9 +1,0 @@
-# Guidelines for model names
-
-The **model** column in the data format should use the commonly used model name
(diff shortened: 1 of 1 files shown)


danielhuppmann:
Sorry, I pushed the wrong commit... Closing in favor of #250

PR #251, opened 2022-12-22
Add region mapping for MEESA v1.2

danielhuppmann:
Per bilateral discussion with @tatar99, we agreed to define a new model MEESA v1.2 to clearly identify that this model has a different spatial resolution than v1.1.

fyi @HauHe @Renato-Rodrigues


tatar99 (approved):

danielhuppmann:
Permissions to submit scenarios for the model "MEESA v1.2" granted for the ECEMF-internal Scenario Explorer for the MEESA modelers team.

HauHe:
Thanks for the heads-up! Is probably the cleanest solution to have a new model version defined.

PR #252, opened 2022-12-22
TIAM-ECN regions and mapping

francescodallalonga:


definitions/region/tiam-ecn_1.2.yaml
```diff
@@ -0,0 +1,37 @@
+- TIAM-ECN 1.2:
+  - TIAM-ECN 1.2|ACE
```
danielhuppmann, line 2: You could use the renaming step to provide more user-friendly names…
```suggestion
  - TIAM-ECN 1.2|<some more descriptive name>
```

danielhuppmann (approved):
Thanks @francescodallalonga, merging now - let's revisit in early 2023 whether we can use more descriptive region names in the ensemble.

PR #256, opened 2023-02-10
Add hydrogen as storage option

erikfilias:
Within case study 3 in openENTRANCE, we are modeling the hydrogen systems with short-term capacity.
In this sense, it would suit us to have hydrogen as a storage option in order to use all the variables related to storage (discharges, injected, etc.) and exchange them between our models.


erikfilias:
I close this PR since its changes (here) are being covered by #291.

PR #257, opened 2023-02-15
add EMF-ECEMF scenario

JohannesEmm:
Adding the EMF-ECEMF comparison collaboration scenario name


danielhuppmann:
fyi @gsaabogado

PR #260, opened 2023-03-16
Add NOAM to the PROMETHEUS region definitions and mapping

danielhuppmann:
Per request by @fragkos1985


fragkos1985 (commented):
I think the change is fine Daniel

PR #262, opened 2023-03-29
Add combined variable for space and water heating

robinhasse:
This adds a new variable equal to the sum of space heating and water heating variables. The same exists already in Ariadne. Would be great if @danielhuppmann or @phackstock could take a look. Its quite urgently needed for ECEMF T2.2 work. Thanks in any case :)

### DIFF
--- a/definitions/variable/energy/energy-final.yaml
+++ b/definitions/variable/energy/energy-final.yaml
@@ -106,6 +106,18 @@
     energy_use: all excl. losses
     sector: Residential and Commercial
     unit: [EJ/yr, GWh]
+- Final Energy|Residential and Commercial|Space and Water Heating:
+    description: Final energy consumed for space and water heating in residential and commercial buildings
+    energy_source: all
+    energy_use: all excl. losses
+    sector: Residential and Commercial
+    unit: EJ/yr
+- Final Energy|Residential and Commercial|Space and Water Heating|{Final Energy carrier}:
+    description: Final energy consumed for space and water heating in residential and commercial buildings
+      from {Final Energy carrier}
+    energy_use: all excl. losses
+    sector: Residential and Commercial
+    unit: EJ/yr
 - Final Energy|Residential and Commercial|Space Heating:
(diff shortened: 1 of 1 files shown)


definitions/variable/energy/energy-final.yaml
```diff
@@ -106,6 +106,18 @@
     energy_use: all excl. losses
     sector: Residential and Commercial
     unit: [EJ/yr, GWh]
+- Final Energy|Residential and Commercial|Space and Water Heating:
+    description: Final energy consumed for space and water heating in residential and commercial buildings
+    energy_source: all
+    energy_use: all excl. losses
+    sector: Residential and Commercial
+    unit: [EJ/yr, GWh]
```
phackstock, line 114: Do you need to allow two different variables?
They also seem to have two different dimensions, Energy/time vs. Energy.

definitions/variable/energy/energy-final.yaml
```diff
@@ -106,6 +106,18 @@
...
+- Final Energy|Residential and Commercial|Space and Water Heating|{Final Energy carrier}:
+    description: Final energy consumed for space and water heating in residential and commercial buildings
+      from {Final Energy carrier}
+    energy_use: all excl. losses
+    sector: Residential and Commercial
+    unit: [EJ/yr, GWh]
```
phackstock, line 120: Same as above

phackstock (commented):
Two small comments in line.

definitions/variable/energy/energy-final.yaml
robinhasse: I kept both units to be consistent with related existing variables. But I go with `EJ/yr` now as for the time being I'll be the only one using it...

phackstock (approved):
Thanks for the quick update @robinhasse. 
Looks good to me.

robinhasse:
Great, thanks for the quick review @phackstock! I don't have rights to write so one of you would need to merge it eventually.

danielhuppmann (approved):
Thanks @robinhasse!

PR #263, opened 2023-04-26
Update scenario names for T5.1

MarkMDekker:
Scenario protocol has been significantly changed, also resulting in different scenarios and scenario names.

### DIFF
--- a/definitions/scenario/scenarios.yaml
+++ b/definitions/scenario/scenarios.yaml
@@ -72,7 +72,7 @@
 # EMF-ECEMF collaboration scenarios
 - EMFECEMF_NZero
 
-# ECEMF - Work package 5 task 5.1 trade limitation scenarios (<= can be removed in the final iteration, but keep it for now)
+# ECEMF - Work package 5 task 5.1 trade limitation scenarios (<= initial scenario names, deprecated)
 - WP5 Default-2C
 - WP5 Default-15C
 - WP5 Short-OECD-2C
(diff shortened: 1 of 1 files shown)


MarkMDekker:
@danielhuppmann  I see one of the checks was unsuccessful. Unsure whether it was really related to these name-changes. Let me know if I need to do anything.

definitions/scenario/scenarios.yaml
```diff
@@ -72,7 +72,7 @@
 # EMF-ECEMF collaboration scenarios
 - EMFECEMF_NZero
 
-# ECEMF - Work package 5 task 5.1 trade limitation scenarios
+# ECEMF - Work package 5 task 5.1 trade limitation scenarios (<= can be removed in the final iteration, but keep it for now)
```
danielhuppmann, line 75: What do you mean here? What is the "final iteration"?

danielhuppmann (commented):
Indeed, this failing test is not related to your PR but some dependency change in our packages. Already investigating...

definitions/scenario/scenarios.yaml
MarkMDekker: Sorry if I'm unclear. We have significantly changed the scenario protocol for T5.1 and also the names and scenario settings are different. I would like to keep having access to the 'old' scenarios that have the 'old' names and that someone that did not upload those 'old' scenarios yet can still do so. Just for a while. I will request a removal of these scenarios when we do not need them anymore (i.e., in the final phase of T5.1, later this year).

danielhuppmann (approved):
Looks good to me, thanks!

danielhuppmann:
Merging in spite of failing tests (not related to this PR, we're investigating the cause)

PR #265, opened 2023-04-26
Fix failing tests in GitHub Actions recipe

danielhuppmann:
There was some change in how `sys.path("..")` was interpreted...


danielhuppmann:
fyi @phackstock

PR #266, opened 2023-04-26
Register mappings for Artelys Crystal Super Grid 2022.1

danielhuppmann:


danielhuppmann:
Permissions added for Artelys Modelers to submit scenarios for the "Artelys Crystal Super Grid 2022.1" model to the ECEMF-internal instance.

PR #267, opened 2023-04-26
add approximate EU regions to region mapping

robertpietzcker:
Add approximate EU regions "EU27 (*)" and "EU27 & UK (*)" to REMIND region mapping

### DIFF
--- a/mappings/remind_2.1.yaml
+++ b/mappings/remind_2.1.yaml
@@ -25,6 +25,10 @@ native_regions:
   - EU27
   - World
 common_regions:
+  - EU27 & UK (*): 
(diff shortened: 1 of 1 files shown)


robertpietzcker:
@danielhuppmann funny that the two asterisks are directly taken as formatting :-)
let's see if I can escape this by "\"

Add approximate EU regions "EU27 (\*)" and "EU27 & UK (\*)" to REMIND region mapping

danielhuppmann (approved):
Thanks, looks good!

danielhuppmann:
Indeed, I didn't think about that...

PR #268, opened 2023-04-26
update remind_3.0 to add approximate EU regions

robertpietzcker:
Add approximate EU regions "EU27 (\*)" and "EU27 & UK (\*)" to REMIND 3.0 region mapping


danielhuppmann (approved):
Looks good!

PR #269, opened 2023-04-26
Add approximate EU regions for WITCH

JohannesEmm:


danielhuppmann (approved):
Thanks @JohannesEmm

danielhuppmann:
FYI @robertpietzcker

PR #274, opened 2023-05-03
Refactor to use nomenclature v0.9

danielhuppmann:
This PR upgrades the workflow to use nomenclature v0.9 and removes the explicit `requirements.txt`.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -25,6 +25,7 @@ jobs:
     - name: Install dependencies
       run: |
         pip install --upgrade pip
+        pip install pytest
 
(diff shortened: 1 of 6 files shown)


phackstock (commented):
Looks good to me. 
One suggestion would be to add pytest as an optional dependency in setup.cfg, then you can just `pip install .[tests]` or something along those lines in the GitHub action pytest.

danielhuppmann:
> One suggestion would be to add pytest as an optional dependency in setup.cfg

Done. @phackstock, can you merge whenever it makes sense for you to switch the container?

phackstock:
Already switched the container, so merging now.

PR #275, opened 2023-05-03
Adding EU27 & UK (*) for OSeMBE

EmiFej:
Added "EU27 & UK (*)" under "common regions". This region is the same as our "EU27 & UK" but the new "EU27 & UK (*)" region is added so that a filter for "EU27 & UK (*)" can be applied, and by doing so getting results from all models.

### DIFF
--- a/mappings/eu-countries.yaml
+++ b/mappings/eu-countries.yaml
@@ -96,6 +96,35 @@ common_regions:
     - Sweden
     - The Netherlands
     - United Kingdom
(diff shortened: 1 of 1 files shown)


danielhuppmann:
Looks good, but please also add the approximate-region `EU27 (*)`.

cc @robertpietzcker

EmiFej:
Thanks for pointing it out @danielhuppmann , now the EU27 (*) has also been added.

PR #276, opened 2023-05-04
Update message_1.2.yaml file for European regions

behnam-zakeri:


danielhuppmann:
@behnam-zakeri, maybe also add the approximate region `EU27 (*)`?

cc @robertpietzcker

robertpietzcker:
Actually, I would say not. 
I specified to the teams that I would prefer them to only add the "EU27 (*)" if they can exclude UK somehow from their results. 
otherwise I think it is best if they only add "EU27 & UK (*)", which would be the level at which ALL teams should report some values

danielhuppmann:
Ok, thanks for the clarification, will merge as is.

PR #278, opened 2023-05-05
Create seasons.yaml

sandrinecharousset:
Description of seasons : this is used to separate a value (eg yearly energy) per season (eg. energy used in winter....) The dates of seasons changing was chosen arbitrarily to the 21st as this is minimising the diffeence of length between seasons, which if sticking to the official calendar is usually bigger


danielhuppmann (commented):
Looks good to me, thanks! Has this been discussed within the relevant case studies? In other words, does this still need to be discussed or can I go ahead and merge the PR)?

sandrinecharousset:
Just added a new yaml file for representative time slices (here representative or average weeks per seasons)..... Would be gratefull to get some comments / advices about it as I am 99% sure there are better ways to write this..... 
@danielhuppmann @erikfilias

definitions/subannual/representative_time_slices.yaml
```diff
@@ -0,0 +1,38 @@
+# description of represnetative periods
```
danielhuppmann, line 1:
```suggestion
# description of representative periods
```

definitions/subannual/representative_time_slices.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+- Average week:
+    duration: 168 hours
+    description: average week over a yearly period for the given variable ie. average value of the 52 weekly values of the variable
+
+- Average week|{Season Input}:
+    duration: 168 hours
+    description: average value of the weekly values of the variable, for the weeks belonging to <season>, which are defined as the weeks with at least 4 days belonging to the considered season
+    
+- hhh|Average week:
```
danielhuppmann, line 11: What does `hhh` stand for? What is the difference to the category `Average week`?

definitions/subannual/representative_time_slices.yaml
```diff
@@ -0,0 +1,38 @@
...
+    note: a week starts on Monday and ends on Sunday
+    
+- Representative week:
+    duration: 168 hours
+    description: Representative week over a yearly period for the given variable ie. value of one of the 52 weekly values of the variable, chosen by the data owner
+    note: Representative Week|<criteria> can be defined, eg. with criteria=holiday
+
+- Representative week|{Season Input}:
+    duration: 168 hours
+    description: Representative week over a yearly period, chosen amongst weeks belnging to <season> for the given variable
```
danielhuppmann, line 28: Use `{}` so that the description from the tag-list ist automatically filled in.

```suggestion
    description: Representative week over a yearly period, chosen amongst weeks belnging to {season} for the given variable 
```

danielhuppmann (changes requested):
Thanks @sandrinecharousset, a few comments inline.

definitions/subannual/representative_time_slices.yaml
sandrinecharousset: @danielhuppmann , hhh is defined in the description section just below, it is meant to be the indexes of each hour of the week. The difference between hhh|Average week and Average week is that Average week is used for the aggregation over the 168 hours of the week (eg the energy produced in the 'average' week) while hhh|Average week is the energy produced at the 'hhh'th  hour of the average week. There are probably better ways of expressing this but did not find another idea......

definitions/subannual/tag_time_input_types.yaml
```diff
@@ -0,0 +1,10 @@
+- Season Input:
+  - winter:
```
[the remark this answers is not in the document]
sandrinecharousset: the reason why I did not is that it didn't work as eg Winter is already used to define a season. I will try to change the definition of seasons to {Input season} and capitalise the list of seasons in Season input.....

danielhuppmann:
I cherry-picked the suggested definitions for the seasons so that at least data at this resolution can be submitted (see #279).

I still do not understand what `hhh` is supposed to mean and how this is supposed to work. Can you please send me a small data file in the IAMC/openENTRANCE format that has this resolution?

sandrinecharousset:
@danielhuppmann , I tried to change the seasons.yaml to avoid double use of Winter , Autumn...... As seasons are described as a 'tag', I wanted to describe the seasons.yaml as {Season Input} but it doesn't pass the validation..... Is it forbidden to use a {} as first part of a name?

definitions/subannual/seasons.yaml
```diff
@@ -0,0 +1,5 @@
+# List of seasons
+# the seasons changing dates was chosen here as to minimise the difference of length between seasons as the objective is to compare energy amounts per seasons
+
+- {Season Input}:
```
danielhuppmann, line 4: You have to use quotes if the first character is a special character like `{`...
```suggestion
- "{Season Input}":
```

erikfilias (approved):

definitions/subannual/tag_time_input_types.yaml
```diff
@@ -0,0 +1,189 @@
...
+  - Summer:
+      duration: 92 / 365
+      description: period between June 21 (included) and Sept. 20 (included)
+
+  - Autumn:
+      duration: 91 / 365
+      description: period between Sept 21 (included) and Dec. 20 (included)
+      
+# Hourly counter within one week 
+- Week Hour Index Input:
```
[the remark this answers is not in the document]
sandrinecharousset: @erikfilias what do you mean by label? 
What I need is to be able to describe representative weeks and average weeks; this means that instead of uploading data for 52 weeks (thus 52*168 rows as my data are hourly), I will only upload 4*168 rows (also because this is more representative and this is how we have presented eg the marginal costs in the report). So I need to express that a row corresponds to eg. the 15th hour of the average winter week. Which is not sth I can write in a format like 2050-01-12 12:00 ..... 
So if you have any idea how to better do that I will be very gratefull (thought of it for days but did not find anything better.....)

PR #279, opened 2023-05-09
Add subannual time resolution for seasons

danielhuppmann:


danielhuppmann:
This PR cherry-picks the four seasons for sub-annual resolution from #278.

PR #280, opened 2023-05-10
Add discharged/Injected Energy for residential load shifting assets

sandrinecharousset:
load shifting assets are 'equivalent' to short term storages but the variables were not defined
These variables are required for uploading openentrance CS1 results

### DIFF
--- a/definitions/variable/technology/electricity-operation.yaml
+++ b/definitions/variable/technology/electricity-operation.yaml
@@ -26,7 +26,7 @@
       of Active Power
     unit: GWh
 - Active Power|Electricity|Load Shifting|{Residential Load Shifting Asset}:
-    description: Energy shifted by assets of kind {Residential Load Shifting Asset} - from the system point of view is equivalent to generation
+    description: Energy shifted by assets of kind {Residential Load Shifting Asset} - from the system point of view is equivalent to generation of active power
       of Active Power
     unit: GWh
 - Active Power|Electricity|{Electricity Storage Type}:
--- a/definitions/variable/technology/electricity-operation.yaml
+++ b/definitions/variable/technology/electricity-operation.yaml
@@ -15,31 +15,31 @@
 
 - Active Power|Electricity|{Electricity Input}:
     description: Active power of {Electricity Input} power plants
-    unit: GWh
+    unit: [MWh, GWh]
 
 - Active Power|Electricity|Load Curtailment:
     description: Energy curtailed - from the system point of view is equivalent to
       generation of Active Power
-    unit: GWh
+    unit: [MWh, GWh]
 - Active Power|Electricity|Load Shifting:
     description: Energy shifted - from the system point of view is equivalent to generation
       of Active Power
-    unit: GWh
+    unit: [MWh, GWh]
 - Active Power|Electricity|Load Shifting|{Residential Load Shifting Asset}:
     description: Energy shifted by assets of kind {Residential Load Shifting Asset} - from the system point of view is equivalent to generation of active power
-    unit: GWh
+    unit: [MWh, GWh]
 - Active Power|Electricity|{Electricity Storage Type}:
     description: Active Power of a {Electricity Storage Type} unit
-    unit: GWh
+    unit: [MWh, GWh]
 - Storage|Electricity|{Electricity Storage Type}:
     description: Available storage of a {Electricity Storage Type} unit
-    unit: GWh
+    unit: [MWh, GWh]
 - Storage|Electricity|Load Shifting|{Residential Load Shifting Asset}:
     description: Available (virtual) storage of a {Residential Load Shifting Asset} unit
-    unit: GWh
+    unit: [MWh, GWh]
 - Storage|Electricity|Load Curtailment:
     description: remaining amount of curtailement
-    unit: GWh
+    unit: [MWh, GWh]
 
 - Spillage|Electricity|{Electricity Storage Type}:
     description: Energy spillage that could not be stored by a {Electricity Storage Type} unit. Several energy models include spillage variables related to renewable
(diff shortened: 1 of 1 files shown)


definitions/variable/technology/electricity-operation.yaml
```diff
@@ -25,12 +25,19 @@
     description: Energy shifted - from the system point of view is equivalent to generation
       of Active Power
     unit: GWh
+- Active Power|Electricity|Load Shifting|{Residential Load Shifting Asset}:
+    description: Energy shifted by assets of kind {Residential Load Shifting Asset} - from the system point of view is equivalent to generation
```
erikfilias, line 29:
```suggestion
    description: Energy shifted by assets of kind {Residential Load Shifting Asset} - from the system point of view is equivalent to generation of active power
```

erikfilias, line 29:
```suggestion
    description: Energy shifted by assets of kind {Residential Load Shifting Asset} - from the system point of view is equivalent to generation of active power
```

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -25,12 +25,19 @@
...
+      of Active Power
```
erikfilias, line 30:
```suggestion
```

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -112,15 +119,35 @@
...
-    description: Amount of energy charged to the storage during the period
-    notes: this represents the total energy which is discharged from the storage during the period, the energy which is charged into the storage during 
+    description: Amount of energy discharged from the storage to the grid during the period
+    notes: this represents the total energy which is discharged from the storage to the grid during the period, the energy which is charged into the storage during 
      that same period is not substracted from this total. This means that the Net Discharged Energy can be computed as Discharged Energy - Charged Energy,
      if Discharged Energy > Charged Energy, 0 if not
-
     unit: GWh
+- Injected Energy|Electricity|Load Shifting|{Residential Load Shifting Asset}:
+    description: Surplus of energy consumed by the {Residential Load Shifting Asset} during the period,
```
erikfilias, line 134: Surplus, or is it just the amount of energy consumed?
I only try to homogenize with the term used in "Discharged Energy|..."

sandrinecharousset: yes, it is surplus. The principle is that a load shifting asset, such as a washing machine already has a nominal consumption profile. When activating the load shifting you will either increase or decrease its consumption in a certain period (and decrease resp. increase it in aanother period, so that the consumption remains the same on a given 'window' (eg 12h for washing machine, 2hours for air conditioning....). The load shifting asset is thus modelled as a virtual storage; from the point of view of this "storage" we could write amount of energy consumed but this may be very ambiguous. (you can have a look at CS1 where this modelling is described )

sandrinecharousset:
@erikfilias I reasked you ot revew as I added 2 more changes to the same file and so did not want to create a new PR as I would have gotten conflicts (and I never lnow how to solve them).......

definitions/variable/technology/electricity-operation.yaml
erikfilias: I also agree with it. "Surplus" fits well since it is an increase or decrease from the consumption made or expected.

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -104,23 +110,47 @@
...
+        compared to nominal consumption profile. This means that the <Residential Load Shifting Asset> is advancing some consumption
+        i.e. consuming earlier than in the nominal profile, or that the consumption had been lowered in an earlier period and is
+        consumed later (delayed)
+    notes: this is equivalent to modelling the <Residential Load Shifting Asset> as a virtual storage 
+      this represents the total energy which is charged into the storage during the period, the energy which is discharged during 
+     that same period is not substracted from this total. This means that the Net Charged Energy can be computed as Charged Energy - Discharged Energy
+     if Charged Energy > Discharged Energy, 0 if not
+    unit: [MWh, GWh]
+- Discharged Energy|Electricity|Load Shifting|{Residential Load Shifting Asset}:
+    description: Amount of energy not consumed by the {Residential Load Shifting Asset} during the period,
```
erikfilias, line 147: I think this variable is contrary to the "Injected...", but I think it is not totally intuitive. The discharged energy is implicit or a consequence of reducing consumption. Perhaps renaming both variables as increasing/decreasing load could be more intuitive. What is your view?

erikfilias (approved):

definitions/variable/technology/electricity-operation.yaml
sandrinecharousset: @erikfilias I did not want to introduce new variable names and stick to those already used for the short term storages.... as it may help treatments of results....  
So What we could do is both keep these variables as they are (they are intuitive for people for whom it is intuitive to model load shifting as short term storages), and also add 2 new variables which could be 
- Increased consumed energy|Electricity|Load Shifting|{Residential Load Shifting Asset}
- Decreased consumed energy|Electricity|Load Shifting|{Residential Load Shifting Asset}
What do you think?
@danielhuppmann any opinion?

danielhuppmann: I advise against having two variables that are basically synonymous, this will just create confusion

danielhuppmann (commented):
Is there a strong reason to allow multiple units for so many variables? In my experience this just creates a headache for users when they want to analyze results and have to bother getting all data for a particular variable into the same unit...

definitions/variable/technology/electricity-operation.yaml
sandrinecharousset: Thanks Daniel, I agree also ; I suggest to keep this pr as it is 
Erik has approved

sandrinecharousset:
> Is there a strong reason to allow multiple units for so many variables? In my experience this just creates a headache for users when they want to analyze results and have to bother getting all data for a particular variable into the same unit...

I would say yes : we are allowing different communities to use them and I have experienced headaches analysing data which were in a unit I had never heard of (Ej/yr)…. In the power system community energies are expressed in GWh ; MWh and GWh are very easy to play with so will not give headaches …… We could have kept only one of the 2, anyway this is not a change from the current pr and was accepted months ago….

danielhuppmann (approved):

PR #281, opened 2023-05-10
Add region definitions and mappings for EMPIRE v1.0

danielhuppmann:
fyi @stianbacke


danielhuppmann:
Permissions assigned for EMPIRE modelers to submit scenarios for „EMPIRE v1.0“ to the openentrance-internal Scenario Explorer

PR #282, opened 2023-05-11
Added available secondary energy|electricity and updated units

sandrinecharousset:
-Available Secondary Energy|Electricity is required to upload CS5 results. It is equivalent to the Secondary Energy that would been achieved if all assets were producing at max capacity during the period considered. In particular for renewable assets, this cannot be computed by multiplying the capacity by the number of hours of the period, as it also relaies on the load factor (eg during the night PV produces 0). 
- change of units for all Secondary Energy|Electricity : the only unit allowed was EJ/yr : in power system studies, we mostly use GWh. Also Ej/yr is limitating as it implies that the variable can only be used for annual data (not subannual); In openentrance CS1,CS4 and CS5 we are uploading Secondary Energy|Electricity per seasons. We could also need monthly or weekly data. This is why I propose that the possible units are EJ/ty (which is currently used in already uploaded data), and GWh (which is needed for power system studies)

### DIFF
--- a/definitions/variable/energy/energy-secondary.yaml
+++ b/definitions/variable/energy/energy-secondary.yaml
@@ -7,23 +7,33 @@
 
 - Secondary Energy|Electricity:
     description: Total net electricity production
-    unit: EJ/yr
+    unit: [EJ/yr, GWh]
 - Secondary Energy|Electricity|{Electricity Input}:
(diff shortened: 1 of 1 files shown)


erikfilias (approved):

danielhuppmann (approved):
Looks good, thanks!

PR #286, opened 2023-05-12
Added Maximum Storage for Hydro|Reservoir and Hydro|Pumped Storage

sandrinecharousset:
When we added the tags for energy storage system, the variable names for storage were redefined with an additionnal section which is |Energy Storage System| This creates an inconsistency in variable names for hydro power: we do have Secondary Energy|Hydro|Reservoir and Capacity|Hydro|Reservoir (which are needed to remain unchainged) for instance but when we need Maximum Storage|Hydro|Reservoir does not exist, instead it is - Maximum Storage|Energy Storage System|Hydro|Pumped Storage, which is not consistent, too complicated and useless (as it is obvious that Reservoir and Pumped Storage are storage); I had thought that it could be possible to allow both names but it doesn't work as we have defined Capacity|{electricity input} and capacity|{electricity storage system} which means that a variable would be duplicated. My suggestion is then to remove Reservoir and Pumped Storage from the list of {Energy storage systems} ; Before doing that I added the variable which I needed which are Maximum Storage|Hydro|Reservoir and Maximum Storage|Hydro|Pumped Storage

### DIFF
--- a/definitions/variable/technology/power-plant.yaml
+++ b/definitions/variable/technology/power-plant.yaml
@@ -130,6 +130,12 @@
 - Maximum Storage|Electricity|{Electricity Storage Type}:
     description: Maximum volume of a {Electricity Storage Type} unit expressed in energy
     unit: [MWh, GWh]
+- Maximum Storage|Electricity|Hydro|Reservoir:
+    description: Maximum volume of Hydro Reservoir technology  expressed in energy
+    unit: [MWh, GWh]
+- Maximum Storage|Electricity|Hydro|Pumped Storage:
+    description: Maximum volume of Pumped Storage technology expressed in energy
+    unit: [MWh, GWh]
 - Minimum Storage|Electricity|{Electricity Storage Type}:
     description: Minimum volume of a {Electricity Storage Type} unit expressed in energy
     unit: [MWh, GWh]


erikfilias (approved):
Yeah, the following variable, `Maximum Storage|Electricity|{Electricity Storage Type}`, is already defined and can be combined with `Energy Storage System|Hydro|Pumped Storage` and `Energy Storage System|Hydro|Reservoir`. If I understood well, you suggested omitting the word `Energy Storage System` from the resulting variable. Then, it's okay, but we should try to keep the compactness of the set of variables by using tags as much as we can, I think.

danielhuppmann:
You can never achieve perfect consistency across the variable tree, so I suggest to keep the current structure - otherwise, we will have an inconsistency between difference storage types like
- Maximum Storage|Electricity|Hydro|Reservoir
- Maximum Storage|Electricity|Energy Storage System|Lithium-Ion

If you really want to harmonize, I'd rather suggest to remove the "Energy Storage System" level.

sandrinecharousset:
> You can never achieve perfect consistency across the variable tree, so I suggest to keep the current structure - otherwise, we will have an inconsistency between difference storage types like
> 
> * Maximum Storage|Electricity|Hydro|Reservoir
> * Maximum Storage|Electricity|Energy Storage System|Lithium-Ion
> 
> If you really want to harmonize, I'd rather suggest to remove the "Energy Storage System" level.

Yes I agree, will propose a PR achieving this and see if @erikfilias can cope with it. The Energy storage system level does to my opinion add nothing apart from complexification

erikfilias:
> > You can never achieve perfect consistency across the variable tree, so I suggest to keep the current structure - otherwise, we will have an inconsistency between difference storage types like
> > 
> > * Maximum Storage|Electricity|Hydro|Reservoir
> > * Maximum Storage|Electricity|Energy Storage System|Lithium-Ion
> > 
> > If you really want to harmonize, I'd rather suggest to remove the "Energy Storage System" level.
> 
> Yes I agree, will propose a PR achieving this and see if @erikfilias can cope with it. The Energy storage system level does to my opinion add nothing apart from complexification

Yeah, I also agree about removing it.

sandrinecharousset:
Included in #291

PR #287, opened 2023-05-15
Added interregional electricity flows

amosschle:
Interregional electricity flows are required to upload CS7 results (Danish electricity imports from and exports to neighbour countries). As far as I understand, the current `Import|Electricity` and `Export|Electricity` do not allow specifying two regions.

### DIFF
--- a/definitions/variable/technology/electricity-operation.yaml
+++ b/definitions/variable/technology/electricity-operation.yaml
@@ -110,7 +110,7 @@
     unit: MWh
 - Flow|Electricity:
     description: Flow from electricity from the former to latter region
-    unit: MWh
+    unit: [MWh, GWh]
 - Unserved Demand|Electricity:
     description: Not served demand
     unit: MWh
--- a/definitions/variable/technology/electricity-operation.yaml
+++ b/definitions/variable/technology/electricity-operation.yaml
@@ -108,7 +108,7 @@
 - Import|Electricity:
     description: Exported electricity from a given region
     unit: MWh
-- Flow|Electricity|{Region}|{Region}:
+- Flow|Electricity:
     description: Flow from electricity from the former to latter region
     unit: MWh
 - Unserved Demand|Electricity:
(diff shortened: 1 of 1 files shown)


definitions/variable/technology/electricity-operation.yaml
```diff
@@ -108,6 +108,9 @@
 - Import|Electricity:
     description: Exported electricity from a given region
     unit: MWh
+- Flow|Electricity|{Region}|{Region}:
```
sandrinecharousset, line 111:
```suggestion
- Flow|Electricity:
```

sandrinecharousset: the column region is meant for describing the region this variable applies to. Directionnal regions are already included and have to be written region1 > region2 (in the region column), which here means that the energy flow is from region1 to region2 (can be negative which means the flow is from region2 to region1)

sandrinecharousset (changes requested):

definitions/variable/technology/electricity-operation.yaml
amosschle: Thanks for the quick review and the info, I didn't know about directional regions. I'll commit your suggestion.

definitions/variable/technology/electricity-operation.yaml
```diff
@@ -108,6 +108,9 @@
 - Import|Electricity:
     description: Exported electricity from a given region
     unit: MWh
+- Flow|Electricity:
+    description: Flow from electricity from the former to latter region
+    unit: MWh
```
sandrinecharousset, line 113:
```suggestion
    unit: [MWh, GWh]
```

sandrinecharousset (changes requested):
propose to allow both MWh and GWh as unit, as this variable will represent an 'energy' flow, and energies are often expressed in GWh

danielhuppmann (approved):
Thanks @amosschle

danielhuppmann:
Tests reminded me that this variable was already added by #280, see [here](https://github.com/openENTRANCE/openentrance/blame/main/definitions/variable/technology/electricity-operation.yaml#L120)

PR #288, opened 2023-05-15
Added max. heat storage

amosschle:
CS7 results include maximum heat storage (MWh). If I understand correctly, only electricity storage is supported by the nomenclature, currently.

### DIFF
--- a/definitions/variable/tag_storage_types.yaml
+++ b/definitions/variable/tag_storage_types.yaml
@@ -29,3 +29,8 @@
       description: redox flow batteries
   - Energy Storage System|Compressed Air:
       description: compressed air storage systems
+
+- Heat Storage Type:
+
+  - Energy Storage System|Heat:
+      description: all types of heat storage systems
--- a/definitions/variable/technology/technologies.yaml
+++ b/definitions/variable/technology/technologies.yaml
@@ -25,7 +25,7 @@
     description: Total installed heat generation capacity for the industrial sector
       from {Fuel}
     unit: GW
-- Maximum Storage|Heat:
+- Maximum Storage|{Heat Storage Type}:
     description: Maximum volume of heat storage expressed in energy
     unit: [MWh, GWh]
 - Capital Cost|Electricity|{Electricity Input}:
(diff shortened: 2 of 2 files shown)


sandrinecharousset:
@amosschle heat storage could be included in the diverse storage systems that were defined in l  definitions/variable/tag_electricity_storage_types.yaml  ; As you are not referring to an electricity storage I would suggest to rename tag_electricity_storage_types.yaml  as tag_storage_types.yaml , and add a new section - Heat Storage Type: in which you could define - Energy Storage System|Heat ; then in technology.yaml you could define Max Storage|{Heat Storage Type} which would create the variable Max Storage|Energy Storage System|Heat  and give also the possibility to further create new variables Max Storage|Energy Storage System|Heat|system1 ..... (and naturally Max Storage|Energy Storage System|Heat=sum of all Max Storage|Energy Storage System|Heat|systemi
@erikfilias @danielhuppmann any idea?

erikfilias:
> @amosschle heat storage could be included in the diverse storage systems that were defined in l definitions/variable/tag_electricity_storage_types.yaml ; As you are not referring to an electricity storage I would suggest to rename tag_electricity_storage_types.yaml as tag_storage_types.yaml , and add a new section - Heat Storage Type: in which you could define - Energy Storage System|Heat ; then in technology.yaml you could define Max Storage|{Heat Storage Type} which would create the variable Max Storage|Energy Storage System|Heat and give also the possibility to further create new variables Max Storage|Energy Storage System|Heat|system1 ..... (and naturally Max Storage|Energy Storage System|Heat=sum of all Max Storage|Energy Storage System|Heat|systemi @erikfilias @danielhuppmann any idea?

I also agree with @sandrinecharousset about renaming the tag related to storage type and including these variables there.

amosschle:
Thanks for the quick review, @sandrinecharousset and @erikfilias. I've made the requested changes.

sandrinecharousset (approved):

danielhuppmann:
There can be only one tag-list per file, please revert the name change of the electricity-storage-type-file and put the heat-storage-types in a separate file.

sandrinecharousset:
included in #291

sandrinecharousset:
included in #291

PR #299, opened 2023-06-14
Eu countries mapping

EmiFej:
Added mapping regions for OSeMBE. The commits were made earlier but no PR was made.


danielhuppmann (approved):
Looks good, thanks!

PR #300, opened 2023-06-15
Update limes_v2.38.yaml - add EU regions with (\*)

robertpietzcker:


danielhuppmann (approved):
Thanks @robertpietzcker!

PR #301, opened 2023-06-19
Add Maximum Storage|{Storage Type}

amosschle:
Following the suggestion by @sandrinecharousset This PR allows for Maximum Storage|Heat, which is an output of CS 7.

### DIFF
--- a/definitions/variable/technology/power-plant.yaml
+++ b/definitions/variable/technology/power-plant.yaml
@@ -124,8 +124,11 @@
       from {Electricity Input}
     unit: MW/h
     skip-region-aggregation: true
-- Maximum Storage|Electricity|{Storage Type}:
+- Maximum Storage|{Storage Type}:
     description: Maximum volume of a {Storage Type} unit expressed in energy
+    unit: [MWh, GWh]
+- Maximum Storage|Electricity|{Storage Type}:
+    description: Maximum volume of an electricity {Storage Type} unit expressed in energy
     unit: [MWh, GWh]
 - Maximum Storage|Electricity|Hydro|Reservoir:
     description: Maximum volume of hydro reservoir expressed in energy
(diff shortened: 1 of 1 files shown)


danielhuppmann:
No, this PR does not add a variable "Maximum Storage|Heat". Please explain what you intend to do (and why).

sandrinecharousset:
> No, this PR does not add a variable "Maximum Storage|Heat". Please explain what you intend to do (and why).

@danielhuppmann , it adds Maximum Storage|{Storage Type}: , and as Heat is included in {Storage Type}, it adds Maximum Storage|Heat (and other variables)

It is a bit difficult to add this variable (which is needed and which needs to be different as the already existing Maximum Storage|Electricity|Heat without changing everything (which we cant as lots of data are already made public)

Maximum Storage|€lectricity|{Storage Type} ismeant or representing electrcity storage (which also include 'virtual electricity storage, eg Maximum Storage|Electricity|Hydro|Reservoir  ; Here what is really stored is water (m3) but what is seen by the models is electricity (GWh) ; We have the same thing for heat => Maximum Storage|Electricity|Heat represents an amount of (virtually) stored electricity , which is stored as heat , while Maximum Storage|Heat represents a real storage of heat. The same amount of heat (in a physical unit for heat) would not lead to the same amount of energy (in EJ or GWh) due to efficiency

We discussed with Amos and I proposed to include this Maximum Storage|{storage type} as it was simple, would not break everyhting , and was not completely studîd (but it should not appear in power-plants.yaml, maybe in technologies.yaml would be more relevant), and also I thought it could start a discussion towards maybe a better idea :-)

Another way ould be

danielhuppmann:
@amosschle, can you please refactor the PR (or close this and start a new one) to simply add a variable "Maximum Storage|Head" and explain in a note how that variable is to be interpreted and used?

amosschle:
> @amosschle, can you please refactor the PR (or close this and start a new one) to simply add a variable "Maximum Storage|Head" and explain in a note how that variable is to be interpreted and used?

Sorry for this coming late. I hope the latest commit works.

danielhuppmann (approved):
Thanks @amosschle!

PR #302, opened 2023-06-19
Update energy-final.yaml

sandrinecharousset:
add missing variable Final Energy|Electricity|Other (excl. Heat, Transport)


danielhuppmann (approved):
Thanks @sandrinecharousset, looks good!

PR #303, opened 2023-06-26
Economy input vars

paopis:
Included new variables in the economy.yaml file to be able to upload the input data that is used in the macroeconomic models to characterize the different scenarios.


definitions/variable/economy/economy.yaml
```diff
@@ -212,4 +212,38 @@
     unit: [EUR_2020/t CO2, USD_2010/t CO2]
     skip-region-aggregation: true
 
-
+- Extraction Limit:
```
danielhuppmann, line 215: Does it make sense to have one limit across all fossil resources? Shouldn't this be distinguished between different fuels?

paopis: In general, it is true. Nevertheless, we have used a single parameter to control the extraction of oil and gas. I can make three parameters: one for oil and one for gas and use the same parameter in the data upload, though. I will make a new pull request with the updated variables

danielhuppmann: No, do NOT make a new pull request - this is not how GitHub works.

paopis: Ok. I only updated the description of the variables, then.

definitions/variable/economy/economy.yaml
```diff
@@ -212,4 +212,43 @@
     unit: [EUR_2020/t CO2, USD_2010/t CO2]
     skip-region-aggregation: true
 
+- Extraction Limit for Natural Gas:
```
danielhuppmann, line 215:
```suggestion
- Extraction Limit|Natural Gas:
```

definitions/variable/economy/economy.yaml
```diff
@@ -212,4 +212,43 @@
...
+    description: Level of allowed extractions for natural gas, relative to year 2011
+    note: The index equals 1 in 2011, the reference year
+    unit: index
 
+- Extraction Limit for Crude Oil:
```
danielhuppmann, line 220:
```suggestion
- Extraction Limit|Crude Oil:
```

danielhuppmann (commented):
Two suggestions to use the hierarchy-notation.

danielhuppmann (approved):
Thanks @paopis!

PR #304, opened 2023-07-03
Add weeks definitions

danielhuppmann:
Add definitions for a weekly subannual resolution


danielhuppmann:
Failing because of a dependency conflict between **nomenclature** and **pydantic** v2.0. Merging anyway.
fyi @phackstock

PR #306, opened 2023-09-07
add missing WP5p1 scenarios

Renato-Rodrigues:
Adding EastWest Tech scenarios defined in ECEMF WP5.1 protocol document, version 1.5.


MarkMDekker (approved):
Yes, this is correct

danielhuppmann:
The test are currently failing because we need to pin some dependencies in this repo, but the new scenarios are ok.

fyi @phackstock

PR #307, opened 2023-09-07
Additional capacity variables

Renato-Rodrigues:
- additional capacity variables support for results in ECEMF scenarios


danielhuppmann (approved):
Looks good to me.

PR #311, opened 2023-09-26
Updates to REMIND regions mapping

Renato-Rodrigues:
Updating REMIND region mapping files used in the ECEMF project to be in line with improvements made in the `eu-climate-advisory-board-workflow` (https://github.com/danielhuppmann/eu-climate-advisory-board-workflow/blob/369de812a08057510fce153887be57f49b9a73e2/mappings/REMIND_3.2.yml), allowing automatic regions aggregation and improving H12 results uploads.

## Changes:

- Including support for REMIND 3.2 model.
- Calculating `World` region results using the iiasa framework.
- Calculating `EU27 & UK (*)` region results using the iiasa framework.
- Adding `NEU` region to support results upload and automatic aggregations while using REMIND H12 regions configuration.
- Improving `OECD & EU (R5)` mapping to support H12 regions aggregation.   

## Warning:

- If you upload REMIND-21 run results, you should not upload `EUR` and `NEU` results to avoid double accounting in the automatically calculated `World` and `EU27 & UK (*)` regions. 
- The `World` region should not be included in new uploads as it is automatically calculated by the iiasa backend from now onward.
- The `EU27` region aggregation should be uploaded directly from REMIND model runs.


mappings/remind_2.1.yaml
```diff
@@ -2,6 +2,18 @@ model: REMIND 2.1
...
+  - IND: REMIND 2.1|IND
+  - JPN: REMIND 2.1|JPN
+  - LAM: REMIND 2.1|LAM
+  - MEA: REMIND 2.1|MEA
+  - OAS: REMIND 2.1|OAS
+  - REF: REMIND 2.1|REF
+  - SSA: REMIND 2.1|SSA
+  - USA: REMIND 2.1|USA
+  # 12-region version of REMIND
+  - EUR: EU27 & UK
```
danielhuppmann, line 14: If this is a native region of the 12-region version of REMIND, I suggest the following 
```suggestion
  - EUR: REMIND 2.1|EUR
```
and add a mapping in the common-region section
```
 - EU27 & UK:
     - EUR
```

Otherwise, it could be confusing that results from a 12-region model run will only have 11 `REMIND 2.1|*` regions - better to duplicate the information.

danielhuppmann (approved):
Looks good to me, thanks!

PR #314, opened 2023-10-25
Adding variables for the number of charging stations

jeremydumoulin:
New variables related to the number of charging stations for electric vehicles, defined in the context of the OpenMod4Africa project, task 3.1.

The number of variables requested is deliberately small for the time being, as it is not yet clear if a higher granularity will be required (e.g.: by also adding chargers at home, at work, or by segmenting the type of charger regarding their technical characteristics).

Many thanks for considering this request. Happy for any comments/suggestions.

### DIFF
--- a/definitions/variable/technology/technologies.yaml
+++ b/definitions/variable/technology/technologies.yaml
@@ -205,6 +205,13 @@
     description: Number of sold (new) vehicles with {Transport mode, tech and carrier}
     unit: million
 
+- Stock|Transportation|Charging station:
+    description: Number of charging stations for electric vehicles
+    unit: million
+- Stock|Transportation|Charging station|Public:
+    description: Number of charging stations for electric vehicles installed in public areas
+    unit: million
+
 - Renovation Rate|{Residential/Commercial}|Envelope:
     description: Rate of renovation of the envelope of {Residential/Commercial} buildings
     unit: '%'
--- a/definitions/variable/technology/technologies.yaml
+++ b/definitions/variable/technology/technologies.yaml
@@ -318,16 +318,16 @@
     description: Number of sold (new) vehicles with {Transport mode, tech and carrier}
     unit: million
 
-- Stock|Transportation|Charging Station:
+- Infrastructure|Transportation|Charging Station:
     description: Number of charging stations for electric vehicles
     unit: million
-- Stock|Transportation|Charging Station|Public:
+- Infrastructure|Transportation|Charging Station|Public:
     description: Number of public charging stations for electric vehicles, i.e. accessible to everybody without restriction (e.g. public parking spot)
     unit: million
-- Stock|Transportation|Charging Station|Semi-Public:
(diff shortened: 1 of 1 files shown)


sandrinecharousset:
@jeremydumoulin I would advise you to request some reviewers to review or this PR may wait a long time.  You could eg request @erikfilias @phackstock and myself  (@phackstock is leading this nomenclature work and is the one who can merge, @erikfilias and myself have experienced the process in openentrance )
Best

definitions/variable/technology/technologies.yaml
```diff
@@ -205,6 +205,13 @@
     description: Number of sold (new) vehicles with {Transport mode, tech and carrier}
     unit: million
 
+- Stock|Transportation|Charging Station:
+    description: Number of charging stations for electric vehicles
+    unit: million
+- Stock|Transportation|Charging Station|Public:
```
danielhuppmann, line 211: In my opinion, this variable name is not sufficiently clear whether they are in public areas (how defined?) or owned by public entities (in contrast to those offered by private companies or on private/residential areas).

Also, please create a complete sub-categorization so that public/private/other charging stations  can be reported such that they be summed up to total number of charging stations.

jeremydumoulin:
@sandrinecharousset : thank you for the advice, but it seems to me that I do not have enough rights to request specific reviewers. I should probably ask @phackstock for that.

@danielhuppmann : Many thanks for following up on this PR. I have added private and semi-public stations to complete the sub-categorization, and also updated the variable descriptions for clarity.
Note that the public/semi-public/private classification is a common way of differentiating charging stations according to their accessibility (see for example: [https://link.springer.com/chapter/10.1007/978-3-319-13194-8_4](https://link.springer.com/chapter/10.1007/978-3-319-13194-8_4)), although I agree that in some cases a higher level of detail may indeed be useful.

Best regards

danielhuppmann (commented):
The distinction between public, semi-public and private makes sense to me.

One final thought - does it make sense to name the variable as "Infrastructure|Transportation|Charging Station|...". It seems to be a bit different in nature from a "stock" like vehicles...

jeremydumoulin:
Thanks for the quick feedback.

Good point. I probably do not have enough experience with variable names to know which is the best option here. It seems that having "Infrastructure|Transportation|Charging Station|..." is indeed more relevant. 
It would also be possible, I suppose, to have something very general like "Number|Charging station|..." as for the number of heat pumps ("Number|{Residential/Commercial}|Heat Pumps"), but I prefer your suggestion. I have updated the PR.

Best

danielhuppmann (approved):
Looks good to me, thanks! Will merge if @sandrinecharousset also approves.

phackstock:
@jeremydumoulin, thanks for the PR, just upgraded your rights to triage so you should be able to select reviewers.

sandrinecharousset (approved):

jeremydumoulin:
Thank you all for processing this PR and for your valuable comments.

PR #320, opened 2024-01-05
OnSSET variables

mvittorio-unh:
New variables for OnSSET


definitions/variable/technology/technologies.yaml
```diff
@@ -374,3 +378,72 @@
...
+- Capital Cost|Electricity|Solar PV|Mini Grid:
+    description: The capital cost (per Capacity unit) of a mini Grid PV system.
+    unit: us$/kW
+    skip-region-aggregation: true
+- Capital Cost|Electricity|Wind|Mini Grid:
+    description: The capital cost (per Capacity unit) of a mini Grid wind powered system.
+    unit: us$/kW
+    skip-region-aggregation: true
+- Capital Cost|Electricity|Hydro|Mini Grid:
+      description: The capital cost (per Capacity unit) of a mini Grid hydropower system.
```
phackstock, line 403: This should be indented less

definitions/variable/technology/technologies.yaml
```diff
@@ -374,3 +378,72 @@
...
+      unit: us$/kW
+      skip-region-aggregation: true
+
+- Capital Cost|Batteries:
+    description: Expected hybrid mini‐grid component costs batteries.
+    unit: us$/kWp
```
phackstock, line 409: I am not familiar with the unit `kWp` what does that refer to?

definitions/variable/energy/energy-secondary.yaml
```diff
@@ -34,6 +34,18 @@
     description: electricity losses from long-range high-voltage transmission
     unit: [EJ/yr, GWh]
 
+#ONSSET START
+- Secondary Energy|Electricity|Transmission Losses|Share:
+    description: This value represents the country’s average technical losses on transmission and distribution
+    unit: %
+- Secondary Energy|Electricity|Distance|Maximum:
```
phackstock, line 41: Do you change these values dynamically per year and is this part of the model output of the ONSSET model? 
To me this looks more like model parameters in which case they are typically not included in the result files.

definitions/variable/energy/energy-secondary.yaml
```diff
@@ -34,6 +34,18 @@
     description: electricity losses from long-range high-voltage transmission
     unit: [EJ/yr, GWh]
 
+#ONSSET START
+- Secondary Energy|Electricity|Transmission Losses|Share:
+    description: This value represents the country’s average technical losses on transmission and distribution
+    unit: %
```
phackstock, line 40:
```suggestion
    unit: '%'
```

definitions/variable/energy/energy-services.yaml
```diff
@@ -42,3 +42,12 @@
 - Energy Service|Transportation|Passenger|{Transport}:
     description: Provision of energy services related to passenger {Transport}
     unit: Gpkm/yr
+
+- Energy Service|Night Lights|Intensity|Minimum:
+    description:  The value ranges from 0 to 63 with higher value representing higher light intensity over 
+        the night. The input parameter sets the minimum light intensity value under which the model will consider a settlement as electrified.
+    unit: -
```
phackstock, line 49: Similar comment to the one in energy-services.yaml, is this a model output or a model parameter? 
If it's the former totally fine to keep if it's the latter I don't think it's needed.
In any case for dimensionless quantities we just leave the unit blank"
```suggestion
    unit:
```

definitions/variable/technology/electricity-expansion.yaml
```diff
@@ -25,3 +25,125 @@
...
+    unit: kW
+
+- Capacity|Added|Hybrid|Wind:
+    description: Total predicted capacity added
+    unit: kW
+
+- Capacity|Added|Hydro:
+    description: Total predicted capacity added
+    unit: kW
+
```
phackstock, line 144: Please remove the empty lines

definitions/variable/technology/electricity-expansion.yaml
```diff
@@ -25,3 +25,125 @@
...
+    description: The investment required by expanding the wind power mini grid to reach the electrification target
+      at the end of the study.
+    unit: us$
+
+- Expansion Cost|Total:
+    description: The total investment required by expanding to reach the electrification target
+      at the end of the study.
+    unit: us$
+
+- Population|Connected|Total:
```
phackstock, line 89: I think you can leave out the total and rename the variable to:
```suggestion
- Population|Connected:
```
the total is implied by the hierarchy with the `|` character.

definitions/variable/technology/electricity-expansion.yaml
```diff
@@ -25,3 +25,125 @@
...
+    description: Total predicted populations with electricity
+    unit: people
+
+- Population|Connected|Grid|Existing:
+    description: Total predicted populations with electricity from existing grid
+    unit: people
```
phackstock, line 95: If this is in the millions I would suggest using
```suggestion
    unit: million
```
otherwise since it's a number I would suggest going fully dimensionless, i.e. using `unit: `.

definitions/variable/technology/electricity-grid.yaml
```diff
@@ -5,7 +5,7 @@
     unit: km
 - Network|Electricity|Expansion Cost:
     description: Cost for increasing the line capacity
-    unit: million EUR/GW/km
+    unit: [million EUR/GW/km, us$]
```
phackstock, line 8: Those units are not compatible.

phackstock (changes requested):
Thanks a lot @mvittorio-unh for the very comprehensive pull request. 

A couple of general points from my side:
* The blank lines between variables are not needed. You can remove them for consistency with the rest of the template.
* The variables that contain `Baseline` or `End year` are not necessary in my opinion since we report annual data anyway. If I understand your variable definitions correctly `Baseline` and `End year` would simply be the first and last value in your timeseries.
* Out of convention, the unit for money we typically use is `US$2010` (example: https://github.com/IAMconsortium/legacy-definitions/blob/d55beb19b80368d0676261da678ede296b9b615c/definitions/variable/variable.yaml#L524-L532). This would need to be changed for all your units involving money. You are obviously free to adjust the year for the dollar value that your model uses, e.g. `US$2020`.
* The unit `people` I would suggest either replacing with `unti: million` or going fully dimensionless, i.e. `unit: `. 

Please find some more comments in line below.

phackstock:
@mvittorio-unh, I fixed some of the issues and now we're getting close to finishing this PR. 
Couple of issues to get through but then we should be done

## Re-defined variables

Currently you re-defined 8 variables. This is an issue since a variable must only be defined once. These are the 8 in question:

```console
ValueError: Collected 8 errors:
  1. Duplicate item in variable codelist: Population|Electricity access|Share
  2. Duplicate item in variable codelist: Network|Electricity|Expansion Cost
  3. Duplicate item in variable codelist: Population|Connected|Grid|Existing
  4. Duplicate item in variable codelist: Population|Connected|Grid|New
  5. Duplicate item in variable codelist: Population|Connected|PV
  6. Duplicate item in variable codelist: Population|Connected|Hybrid|PV
  7. Duplicate item in variable codelist: Population|Connected|Hybrid|Wind
  8. Duplicate item in variable codelist: Population|Connected|Hydro
```

I could just remove the ones that you added but at least for some of them you use different units to what was already there. Please take a look to make sure that your model outputs conform to the variables and remove your variables accordingly.

## `End Year` variables

There are still a number of variables that contain `End Year`. From my understanding that would just be the last value in the time series. If that's correct please remove these variables. If my assumption is not correct, just let me know.

## `skip-region-aggregation: true`

There are a number of variables that you introduced where you put `skip-region-aggregation: true`. Is that on purpose? What this attributes means that this variable will be skipped for region aggregation. As we're currently implementing African regions (West Africa, North Africa, etc...) we would want to aggregated as many variables as we can. So if you remove this line our processing will attempt to aggregate the variable.

## Non-sum aggregations

There are a number of variables that are maximum or minimum values. These need special aggregation as the default is to just sum values. I'd be happy to help with that if you provide me with a list of variables that you'd like to add where you would say that a summation is not the right way to compute a meaningful aggregate.

## Open comments

I have closed all the comments that you addressed from my last review. However, there's still a few open ones left. Please have a look, you can answer them in the comments directly.

phackstock:
@mvittorio-unh, thanks for your answers via email. I've addressed all the open points mentioned above.
Unfortunately, we have a new (hopefully final) set of errors:

```console
The following variables are used as 'weight' for aggregation but are not defined in the variable codelist:
'Final Energy|Diesel' used for 'Price|Final Energy|Diesel|Low' in: variable/energy/energy-prices.yaml
'Final Energy|Diesel' used for 'Price|Final Energy|Diesel|High' in: variable/energy/energy-prices.yaml
'Primary Energy|Electricity' used for 'Price|Primary Energy|Electricity' in: variable/energy/energy-prices.yaml
'Primary Energy|Electricity' used for 'Price|Primary Energy|Electricity|Expansion' in: variable/energy/energy-prices.yaml
```

these errors mean that as an aggregation weight you've used a variable that does currently not exist.
There are two possible fixes for this solution:

1. You take a look at the current list of variables and find one that most closely matches the weight that you would want to use
2. You add the missing variables to the codelists

Option 1. is preferable for me since it does not introduce new variables. If you can't find a good fit though adding new variables is totally fine as well.

dc-almeida:
@mvittorio-unh @phackstock removed the variables as per the email, and resolved a merge conflict with the original repo regarding economy.yaml variables (the fork was behind main). Please double-check it's all good there.

definitions/variable/technology/technologies.yaml
```diff
@@ -492,3 +496,60 @@
...
+    description: The capital cost (per Capacity unit) of a mini Grid wind powered system.
+    unit: US$2020/kW
+    skip-region-aggregation: true
+- Capital Cost|Electricity|Hydro|Mini Grid:
+    description: The capital cost (per Capacity unit) of a mini Grid hydropower system.
+    unit: US$2020/kW
+    skip-region-aggregation: true
+- Capital Cost|Batteries:
+    description: Expected hybrid mini-grid component costs batteries.
+    unit: US$2020/kWp
```
phackstock, line 525: You have to set `skip-region-aggregation: true` here. Reason being that the default region aggregation simply adds up all the values for across different regions. For a variable with a unit like `US$2020/kWp` this would lead to wrong values. 
We do offer the possibility of using a weighted average but for now, for simplicity let's skip region aggregation.
The same logic applies for all variables with `US$2020/kWp` and `US$2020/km` as a unit.

phackstock (changes requested):
Region aggregation needs to be skipped for a number of variables. 
See reasoning in-line below.

dc-almeida:
@mvittorio-unh there are some pending requested changes in this PR to merge it. Any help needed?

PR #335, opened 2024-05-01
edit power-plants.yaml

sandrinecharousset:
add variable Emission Rate[CO2 which gives the amount of CO2 emitted when producing 1GWh with each techno


sandrinecharousset:
variable exists already

PR #336, opened 2024-05-01
Update electricity-grid.yaml

sandrinecharousset:
add unit EUR_2020/MW to allow using this variable for extension not only of the line length but also of the line capacity

We could also create a new variable whcih I would have named Network|Electricity[Capital Cost for consistance with the investment costs in other technologies which are named Capcital Cost

### DIFF
--- a/definitions/variable/technology/electricity-grid.yaml
+++ b/definitions/variable/technology/electricity-grid.yaml
@@ -5,7 +5,7 @@
     unit: km
 - Network|Electricity|Expansion Cost:
     description: Cost for increasing the line capacity
-    unit: million EUR/GW/km, EUR_2020/MW
+    unit: [million EUR/GW/km, EUR_2020/MW]
 - Network|Electricity|Capacity:
     description: Capacity of the line
     unit: MW
--- a/definitions/variable/technology/electricity-grid.yaml
+++ b/definitions/variable/technology/electricity-grid.yaml
@@ -5,7 +5,7 @@
     unit: km
 - Network|Electricity|Expansion Cost:
     description: Cost for increasing the line capacity
-    unit: million EUR/GW/km
+    unit: million EUR/GW/km, EUR_2020/MW
 - Network|Electricity|Capacity:
     description: Capacity of the line
     unit: MW


erikfilias:
> add unit EUR_2020/MW to allow using this variable for extension not only of the line length but also of the line capacity
> 
> We could also create a new variable whcih I would have named Network|Electricity[Capital Cost for consistance with the investment costs in other technologies which are named Capcital Cost

Hello, the capital cost variables are in the `technology.yaml` file. I suggest moving these variables to the `electricity-expansion.yaml` file and adding this new variable as `Capital Cost|Electricity|Network`. Finally, I would rename the `electricity-expasion.yaml` file to `invesment.yaml` in order to keep the coherence that this file also comprises investment variables related to heat and hydrogen. @sandrinecharousset , @phackstock , How do you see it?

sandrinecharousset:
@erikfilias sounds good to me
The variable name you propose is more consistent with other variable names 
I am not fully satisfied with having only one capital cost for network expansion in capacity and in length , would be better to separate but could not find a good way of doing it
Regarding moving variables from technologies not sure it is a good idea as lots of other models do use these variables and not those in electricity focused yaml files so I would put the new variable maybe in technologies 
Regarding electricity-expansion from what I remember this is describing variables that are results of electricity capacity expansion models no? No pb to rename it though but I wouldn’t put the capital cost there 
@danielhuppmann ypu may have an opinion there also?
Best

phackstock:
Thanks @sandrinecharousset and @erikfilias for the discussion and the inputs.
@sandrinecharousset I very much agree with you that we should split the variable in two to differentiate between line length and capacity. 
For naming things I always like to err on the side of verbose but immediately obvious rather than trying to be clever but potentially misleading. Therefore I'd propose something like:

* `Capital Cost|Electricity|Network Capacity`
* `Capital Cost|Electricity|Network Length`

or something along those lines.

definitions/variable/technology/electricity-grid.yaml
```diff
@@ -5,7 +5,7 @@
     unit: km
 - Network|Electricity|Expansion Cost:
     description: Cost for increasing the line capacity
-    unit: million EUR/GW/km
+    unit: million EUR/GW/km, EUR_2020/MW
```
danielhuppmann, line 8:
```suggestion
    unit: [million EUR/GW/km, EUR_2020/MW]
```

danielhuppmann (commented):
Thank you for the discussion. One point of caution (re the suggestion by @phackstock): the cost are not "per length", but in "capacity per length", so the suggested variable names can also be misleading.

So for simplicity, I suggest to go with the simple suggestion by @sandrinecharousset (but with a minor formatting fix to make sure that the nomenclature package parses the units correctly).

definitions/variable/technology/electricity-grid.yaml
phackstock: Not sure how this might affect aggregation if we have two variables of the same name but different units. @sandrinecharousset would you report using both units?

danielhuppmann: *pyam* treats data for the same variable with different units as distinct timeseries, so should aggregate them separately.

sandrinecharousset: fine for me, many thanks

sandrinecharousset:
@phackstock could you merge this so that we can upload datasets for the ws with the good unit?

phackstock (approved):

PR #338, opened 2024-05-07
Update tag_storage_types.yaml

sandrinecharousset:
The comment says 'does not include hydro' but Pumped Storage was still included.... which would create variables such as eg Capacity|Electricity|Pumped Storage while the use of Electricity Input creates Capacity|Electricity|Hydro|Pumped storage I propose to remove pumped storage from storage_types


erikfilias (approved):

sandrinecharousset:
@phackstock could you have a look and merge if ok?

phackstock (approved):

PR #343, opened 2024-06-06
Register region-aggregation for Enertile and FLEX

danielhuppmann:


danielhuppmann:
Permissions added for FLEX and ENERTILE modelers to submit scenarios to the ECEMF-internal Scenario Explorer

PR #345, opened 2024-06-18
Update power-plant.yaml

sandrinecharousset:
added lifetime variable


sandrinecharousset:
variable was already there

PR #348, opened 2024-07-16
Update countries.yaml

sandrinecharousset:
changed iso2 code of norway from NO to 'NO' because python yaml understood NO as False, not as a 2 characters string

### DIFF
--- a/definitions/region/countries.yaml
+++ b/definitions/region/countries.yaml
@@ -161,7 +161,7 @@
       iso3_codes: MKD
   - Norway:
       eu_member: false
-      iso2: NO
+      iso2: 'NO'
       iso3: NOR
       iso3_codes: NOR
   - Poland:


sandrinecharousset:
I am using this file in one of my scripts for plan4res and I couldn't get the iso2 code for norway, it was replaced by False...... Maybe there is a better solution but I am not a python nor yaml expert..... @danielhuppmann  @phackstock  you may know a better solution

danielhuppmann (approved):
Right, this is a known issue with yaml - I thought that this had been fixed long ago...

Thanks for the PR!

sandrinecharousset:
> Right, this is a known issue with yaml - I thought that this had been fixed long ago...
> 
> Thanks for the PR!

thanks @danielhuppmann for the quick merge!!

PR #349, opened 2024-09-20
Update energy-final.yaml

sandrinecharousset:
change description of variable Final Energy|Electricity|Heat which was ambiguous (seemed to only include electric heating while it in reality also includes production of industrial heat for industrial processes) + add variable Final Energy|Electricity|Heating which only comprises the electricity used for electric heating of buildings (incl residential, commercial, industrial.....)

We need to separate the share of electricity which is thermo sensitive (more heating when temperature is colder) from the industrial heat (for industrial processes) which is not dependent of the temperature


danielhuppmann (approved):
Thanks @sandrinecharousset

PR #350, opened 2024-09-30
Add region definitions and mappings for REMIND 3.3

danielhuppmann:


danielhuppmann:
FYI @JakobBD @dc-almeida

PR #353, opened 2024-10-07
add non-CO2 emissions in co2e

robertpietzcker:
for easier summation checks and finding of cross-model differences, also include the non-CO2 gases at the highest aggregation level in CO2e metric

### DIFF
--- a/definitions/variable/emissions/emissions.yaml
+++ b/definitions/variable/emissions/emissions.yaml
@@ -678,6 +678,23 @@
     description: Kyoto GHG emissions from other sources, including CO2, CH4, N2O and F-gases 
       (preferably use 100-year GWPs from AR4 for aggregation of different gases)
     unit: Mt CO2e/yr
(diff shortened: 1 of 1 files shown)


robertpietzcker:
@danielhuppmann would it be possible to merge this?

danielhuppmann (approved):
Ok for me to merge here but I don’t think we want to add this in common-definitions…

robertpietzcker:
> Ok for me to merge here but I don’t think we want to add this in common-definitions…

Thanks for merging it here!

why wouldn't you want to have this in common-definitions? I think it is very helpful for plotting, to check consistency of reporting and understand differences between models

danielhuppmann:
Well, `Emissions|Kyoto Gases|CH4` should be identical to `Emissions|CH4` except for the unit, right?
And `Emissions|Kyoto Gases|non-CO2` should be `Emissions|Kyoto Gases` minus `Emissions|CO2`.

So seems to be duplication of existing variables.

PR #354, opened 2024-10-07
add hydrogen SE and FE and gross trade SE

robertpietzcker:
Add gross trade values for electricity

Add liquids/gases from hydrogen to Secondary Energy when hydrogen is modelled as a commodity (see also discussion in https://github.com/IAMconsortium/common-definitions/pull/106#issuecomment-2328867412 )

Also add sub-variables for Curtailment for an analysis on storage/curtailment in ECEMF. 

Also added centralized heat generation by heat pumps

### DIFF
--- a/definitions/variable/energy/energy-secondary.yaml
+++ b/definitions/variable/energy/energy-secondary.yaml
@@ -84,7 +84,7 @@
     description: Total centralized heat generation from {Fuel}
     unit: EJ/yr
 - Secondary Energy|Heat|Electricity|Heat Pumps:
-    description: Total centralized heat generation from heat pumps ("Electricity" is already covered in {Fuel})
+    description: Total centralized heat generation from heat pumps
     unit: EJ/yr
 - Secondary Energy|Heat|Nuclear:
     description: centralized heat generation from nuclear energy
--- a/definitions/variable/energy/energy-secondary.yaml
+++ b/definitions/variable/energy/energy-secondary.yaml
@@ -24,6 +24,22 @@
     description: Curtailment of electricity production due to oversupply from
       variable renewable sources (typically from wind and solar)
     unit: [EJ/yr, GWh]
+- Secondary Energy|Electricity|Curtailment|Solar:
+    description: Curtailment of electricity from solar
+    unit: [EJ/yr, GWh]
+- Secondary Energy|Electricity|Curtailment|Solar|PV:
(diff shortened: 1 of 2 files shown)


robertpietzcker:
@danielhuppmann would it be possible to merge this PR?

robertpietzcker:
@danielhuppmann pushing this up again :-)

definitions/variable/energy/energy-secondary.yaml
```diff
@@ -56,10 +78,13 @@
...
 - Secondary Energy|Heat:
-    description: Total heat generation
+    description: Total centralized heat generation
     unit: EJ/yr
 - Secondary Energy|Heat|{Fuel}:
-    description: Total heat generation from {Fuel}
+    description: Total centralized heat generation from {Fuel}
+    unit: EJ/yr
+- Secondary Energy|Heat|Electricity|Heat Pumps:
+    description: Total centralized heat generation from heat pumps ("Electricity" is already covered in {Fuel})
```
danielhuppmann, line 87: This seems to be a copy-paste error with {Fuel}

robertpietzcker: sorry, I am failing to see the copy-paste error - could you clarify? @danielhuppmann

danielhuppmann: There is a tag `{Fuel}` in the description but no corresponding tag in the variable name - and the whole second part of the description doesn’t make sense to me…

robertpietzcker: I meant to explain why there is an explicit `Electricity|Heat Pumps` version of the `Secondary Energy|Heat|` variable, but no explicit `Electricity` version - that one is already included in the `{Fuel}` version of `Secondary Energy|Heat|`.

But I can easily remove this explanation if you find it self-explanatory

danielhuppmann: I still don’t understand what you want to say. But in any case, please remove the curly brackets because are reserved for tags that are automatically replaced - and they can only be replaced if the same tag is present in the variable name itself.

robertpietzcker: good to know. I just updated it accordingly

danielhuppmann (approved):
Looks good to me, thanks!

PR #360, opened 2024-11-27
Add explicit list of dimensions used in validation

danielhuppmann:


danielhuppmann:
FYI @dc-almeida @phackstock - removing the implicit variable/region default in **nomenclature** made the tests fail in this repo, but adding the dimensions-attribute to nomenclature allows to fix it quite easily.

I'll take a look if there are other workflow-repos where I had added empty folders to definitions. (I had added it here to have a Readme for each IAMC dimension - early days when we didn't have nice documentation in other places).

PR #361, opened 2024-11-27
Add region definitions and mappings for IMAGE 3.4

danielhuppmann:
FYI @VassilisDaioglou


danielhuppmann:
Added permissions for the IMAGE team to submit scenarios for the IMAGE 3.4 model to the ECEMF-internal Scenario Explorer.

PR #363, opened 2025-02-07
add aggregated sales/stock variables

robertpietzcker:
in `{Transport mode, tech and carrier}`, only the detailed categories (eg "Bus|BEV") are not contained, not the aggregated category ("Bus"). 

Therefore, the more aggregated categories were added by adding a "{Transport mode}" entry for Sales/Stock


robertpietzcker:
@danielhuppmann @phackstock is this PR ok, could you merge it?
Thanks!

danielhuppmann (approved):
Thanks for the reminder!

PR #364, opened 2025-02-19
Add total energy system cost variable

robertpietzcker:
To make the "Policy Cost|Additional total energy system cost" less failure-prone, it is better to have the actual total energy system costs explicitly for each scenario, so that the delta can be calculated in the analysis itself. Also, this allows for easier consistency checks between models (how similar are the boundaries of what they count under "energy system costs")


danielhuppmann (approved):
Looks good to me, thanks!

PR #366, opened 2025-03-03
Add more trade variables for secondary energy types

robertpietzcker:
Added secondary energy trade variables for solids and gases. 
Also reformatted other trade variables to also use the "notes" formulation.
Also added the information that amounts reported under "Trade|Secondary Energy" should not additionally be reported under "Trade|Primary Energy|xxx"

### DIFF
--- a/definitions/variable/economy/economy.yaml
+++ b/definitions/variable/economy/economy.yaml
@@ -144,7 +144,7 @@
     skip-region-aggregation: true
 
 - Energy System Cost:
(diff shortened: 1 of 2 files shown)


danielhuppmann (approved):
Thanks! And could you please double-check with common-definitions whether the variables are also included there?

robertpietzcker:
> Thanks! And could you please double-check with common-definitions whether the variables are also included there?

Not yet, but will do so!

PR #368, opened 2025-03-27
Update energy-secondary.yaml for typo in trade of electricity

behnam-zakeri:
"Trade|Gross Imports" should be "Trade|Gross Import" (without "s") to be consistent with the variable list in WP1.


danielhuppmann (approved):
As far as I can tell, only MESSAGE reported this variable, and it was not included in the public ECEMF Scenario Explorer. So I changed the variable-name directly in the internal database for consistency with this PR.

PR #370, opened 2025-04-30
Add production & value addition variables in economy.yaml

awais307:
Added production variables for industry sector in the economy.yaml file.

### DIFF
--- a/definitions/variable/economy/economy.yaml
+++ b/definitions/variable/economy/economy.yaml
@@ -15,6 +15,15 @@
 - Production|Chemicals|Methanol [Volume]:
     description: Production of methanol in physical units
     unit: Mt/yr
+- Production|Steel [Volume]:
+    description: Production of steel in physical units
+    unit: Mt/yr
+- Production|Steel|Primary [Volume]:
+    description: Production of primary steel in physical units
+    unit: Mt/yr
+- Production|Steel|Secondary [Volume]:
+    description: Production of secondary steel in physical units
+    unit: Mt/yr
 - Production|Iron and Steel|Steel [Volume]:
     description: Production of steel in physical units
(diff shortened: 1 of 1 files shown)


definitions/variable/economy/economy.yaml
```diff
@@ -15,6 +15,48 @@
...
+- Production|Chemicals|Ammonia [Volume]:
+    description: Production of ammonia in physical units
+    unit: Mt/yr
+- Production|Chemicals|High-Value Chemicals|Ethylene [Volume]:
+    description: Production of ethylene in physical units
+    unit: Mt/yr
+- Production|Chemicals|Methanol [Volume]:
+    description: Production of methanol in physical units
+    unit: Mt/yr
+- Production|Iron and Steel|Steel|Primary [Volume]:
```
behnam-zakeri, line 27: Shouldn't we remove the previous variables that are no duplicated:
`Production|Steel [Volume]`
`Production|Steel|Primary [Volume]`
`Production|Steel|Secondary [Volume]`

awais307: Thanks for pointing out. I addressed this with the recent commit.

phackstock (approved):
Looks good to me in principle. Just want to point out that variable renaming is always a little bit tricky as we might already have variables with the old names uploaded.

behnam-zakeri:
@awais307 and @phackstock I checked the internal scenario explorer, and it seems some models are indeed using the old-format variable names in WP4. So, we shouldn't remove the old format and just add the new ones.

behnam-zakeri:
@danielhuppmann and @phackstock I believe this is ready for being merged. We didn't remove the existing variables and just added some new ones.

PR #372, opened 2025-06-15
add variables for the different shares of electricity consumption

sandrinecharousset:
additionnal clarification.
Electricity consumption is composed of:
- electric heating
- production of heat for industrial processes
- cooling
- e-transport
- electrolysis
- other uses

### DIFF
--- a/definitions/variable/energy/energy-final.yaml
+++ b/definitions/variable/energy/energy-final.yaml
@@ -513,6 +513,7 @@
     sector: all
     unit: [EJ/yr, GWh]
 
+# the variables Final Energy|Electricity|* define the different shares of the electricity consumption
 - Final Energy|Electricity|Cooling:
     description: Final electricity consumption for cooling,
     energy_source: electricity
@@ -525,8 +526,14 @@
     energy_use: industry and heating
     sector: all
     unit: [EJ/yr, GWh]
+- Final Energy|Electricity|Industry:
+    description: Final electricity consumption for industrial processes,
+    energy_source: electricity
+    energy_use: industry and heating
+    sector: all
+    unit: [EJ/yr, GWh]
 - Final Energy|Electricity|Heating:
-    description: Final electricity consumption forelectric heating),
+    description: Final electricity consumption for electric heating),
     energy_source: electricity
     energy_use: heating
     sector: all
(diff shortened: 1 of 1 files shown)


definitions/variable/energy/energy-final.yaml
```diff
@@ -564,6 +571,34 @@
...
     sector: all
     unit: [EJ/yr, GWh]
+- Final Energy|Electricity|Other (excl. Heating, Industry, Cooling, Transport):
+    description: Final electricity consumption for non heating, non cooling, non e-transport, non industrial heat
+      uses,
+    energy_source: electricity
+    energy_use: all excl.heating, cooling, transportation
+    sector: all
+    unit: [EJ/yr, GWh]
+- Final Energy|Electricity|Other (excl. Heating,Industry, Cooling, Transport, Cooking):
```
phackstock, line 581: Just to keep the convention:
```suggestion
- Final Energy|Electricity|Other (excl. Heating, Industry, Cooling, Transport, Cooking):
```

definitions/variable/energy/energy-final.yaml
```diff
@@ -564,6 +571,34 @@
...
+    sector: all
+    unit: [EJ/yr, GWh]
+- Final Energy|Electricity|Other (excl. Heating, Industry, Transport):
+    description: Final electricity consumption for non heating, non e-transport, non industrial heat
+      uses,
+    energy_source: electricity
+    energy_use: all excl.heating, cooling, transportation
+    sector: all
+    unit: [EJ/yr, GWh]
+- Final Energy|Electricity|Other (excl. Heating,Industry, Transport, Cooking):
```
phackstock, line 595:
```suggestion
- Final Energy|Electricity|Other (excl. Heating, Industry, Transport, Cooking):
```

phackstock (commented):
To aid comparability, I'd suggest to use only one unit of energy, either `EJ/yr` or `GWh`. Also in the case of `GWh` I guess we'd need a `/yr` as well, correct?

sandrinecharousset:
> To aid comparability, I'd suggest to use only one unit of energy, either `EJ/yr` or `GWh`. Also in the case of `GWh` I guess we'd need a `/yr` as well, correct?

@phackstock we already had the discussion about EJ / GWh a few years ago with @danielhuppmann . I would prefer to keep both as the energy modelling community uses EJ while the power modelling community uses GWh and has no clue what EJ is. From this we had concluded to keep both units possible and I would prefer not to change this. (in particular for the variables Final Energy|Electricity* which are power demands and the natural unit is GWh.  Regarding using GWh or GWh/yr , same thing;. I do not quite understand why we are using EJ/yr instead of EJ , and I believe the /yr does not add anything. Also we can uses these variables on subannual timesteps , and then using the /yr would create an error..... My suggestion would rather be to remove all /yr (as when the datafile only has annual columns then it means that it is yearly energies, while when it has subannual columns - we are using those in output files of power models - is is energies on a shorter period (eg we can compute the Final Energy[xxxx on a subannual period which can be eg Spring or Summer, or January

phackstock (approved):
Thanks for the detailed explanation @sandrinecharousset. In this case good to keep the convention. It just adds overhead later on for comparison but that seems to be a good price to pay in this case.

PR #373, opened 2025-06-19
Euro-Calliope native regions and mapping

FraSanvit:
Add two files for region mapping and definitions of the native regions


danielhuppmann:
Closing because the model is already registered, see https://github.com/openENTRANCE/openentrance/blob/eba40f35cde680b6d2f61435754b26480bf4581e/mappings/eu-countries.yaml#L3

